PR #1, opened 2023-04-13
refactor + qol additions + flake-compat

getchoo:
the refactors include:

- using `inherit` with `let in` more

  - this is mainly so things look a bit nicer while also stopping contributors
    from worrying about typing in `lib` or `builtins` a lot

- moving everything to `modules`

  - this follows general praticies from other projects in the nix community where
    modules (both nixos and home-manager) are placed in a dedicated folder

- formatting

  - i don't think a formatter was used on this? i could be wrong, but i used
    alejandra here just in case. i think it would be good to also decide on
    a standard formatter to put in the flake (which would of course be applied
    to this pr)

the main qol of change here is that all files in a modules can now be imported automatically,
meaning contributors won't need to manually add a new file to the `imports`
array in `default.nix`

and finally, flake-compat support has been introduced, giving our non-flake
users a better time interacting with our modules

i've tested the nixos and home-manager modules on nixos unstable, and i've found no
problems 👍🏻


Stonks3141:
Thanks for the PR! It mostly looks good, but I don’t think that using `let inherit …; in` is a win for readability and accessibility to new contributors. Not only is it unclear when to use that vs typing out `lib` (the code includes both), but when using something from `lib` or `builtins` you have to go to the top of the file, add it, and go back. I would rather see something like `options.foo = with lib; ...` instead when there is a lot of `lib`.
As for formatting, I had kind of hoped that we wouldn’t need to add nixpkgs as an input just for that since it will get downloaded by users, I’d like to hear your thoughts on that.
In any case, I will be closing this because its scope is so large and so that some of the other changes can be merged while we discuss style. Could you split this PR into one for flake-compat, one for `modules/`, one for auto-import, and another for the other refactors? Oh, and add yourself to the thanks section in the readme in one of them. 😉

getchoo:
> I would rather see something like options.foo = with lib; ... instead when there is a lot of lib.

noted 

> As for formatting, I had kind of hoped that we wouldn’t need to add nixpkgs as an input just for that since it will get downloaded by users, I’d like to hear your thoughts on that.

i think it could be done through github actions then, or if you are open to making nixpkgs itself an input, we could recommend people to use `inputs.nixpkgs.follows = "nixpkgs";` when making this repository an input.

> In any case, I will be closing this because its scope is so large and so that some of the other changes can be merged while we discuss style. Could you split this PR into one for flake-compat, one for modules/, one for auto-import, and another for the other refactors?

yep! i'll remember to turn off auto format this time 😆

PR #2, opened 2023-04-14
add flake-compat support

getchoo:
continuation of #1, this just adds flake-compat support

sidenote: `{ self }` was removed due to nix pushing `flake-compat` to
the arguments of `output` (which wouldn't work without the `...` syntax)
and it not being used anyways.


Stonks3141:
Thanks!

PR #4, opened 2023-04-14
auto import files in modules

getchoo:
this adds the ability to autoimport files from #1, making contributing a bit easier :)

i also forgot to update the paths for the modules in #3, so i fixed that here as well

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@
   };
 
   outputs = _: {
-    nixosModules.catppuccin = import ./nixos;
-    homeManagerModules.catppuccin = import ./home-manager;
+    nixosModules.catppuccin = import ./modules/nixos;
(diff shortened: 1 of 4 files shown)


Stonks3141:
I looked at this again and I’m not convinced that this actually improves readability or makes things easier for new contributors. The alternative is just adding a file to the `imports` attr, and this adds a lot of rather inscrutable functional code. I would prefer to avoid this kind of magic when possible, as I know I’ve often been confused by magic when contributing to things. I’ll leave this open for now in case anyone else wants to discuss this.

Stonks3141:
It looks like home-manager also has a big import list: https://github.com/nix-community/home-manager/blob/master/modules/modules.nix#L16-L328.

Stonks3141:
After some discussion on Discord, I think I'll close this.

PR #7, opened 2023-04-14
add support for gtk

getchoo:
this adds support for the [gtk](https://github.com/catppuccin/gtk) theme, as well as an accent option for future use in other modules :)

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -31,7 +31,7 @@ in {
       # string -> string
       # this capitalizes the first letter in a string
       # it's used to set the theme name correctly here
-      mkUpper = word: (toUpper (substring 0 1 word)) + (substring 1 999 word);
+      mkUpper = word: (toUpper (substring 0 1 word)) + (substring 1 (stringLength word) word);
 
       flavourUpper = mkUpper cfg.flavour;
       accentUpper = mkUpper cfg.accent;
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -45,7 +45,7 @@ in {
         "Catppuccin-${flavourUpper}-${sizeUpper}-${accentUpper}-${gtkTheme}";
       package = pkgs.catppuccin-gtk.override {
         inherit (cfg) size tweaks;
-        accents = [cfg.accent];
+        accents = [ cfg.accent ];
         variant = cfg.flavour;
       };
     };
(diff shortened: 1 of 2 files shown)


modules/home-manager/gtk.nix
```diff
@@ -0,0 +1,52 @@
...
+    };
+  };
+
+  config.gtk.theme = with lib;
+    with builtins;
+    let
+      # string -> string
+      # this capitalizes the first letter in a string
+      # it's used to set the theme name correctly here
+      mkUpper = word: (toUpper (substring 0 1 word)) + (substring 1 999 word);
```
Stonks3141, line 34:
```suggestion
      mkUpper = word: (toUpper (substring 0 1 word)) + (substring 1 (stringLength word - 1) word);
```

modules/home-manager/gtk.nix
```diff
@@ -0,0 +1,52 @@
...
+      description = "Catppuccin accents for gtk";
+    };
+    size = mkOption {
+      type = types.enum [ "standard" "compact" ];
+      default = "standard";
+      description = "Catppuccin size variant for gtk";
+    };
+    tweaks = mkOption {
+      type = types.listOf (types.enum [ "black" "rimless" "normal" ]);
+      default = ["normal"];
```
Stonks3141, line 23:
```suggestion
      default = [ "normal" ];
```

modules/home-manager/gtk.nix
```diff
@@ -0,0 +1,52 @@
...
+
+      # use the light gtk theme for latte
+      gtkTheme = if cfg.flavour == "latte" then "Light" else "Dark";
+
+    in mkIf cfg.enable {
+      name =
+        "Catppuccin-${flavourUpper}-${sizeUpper}-${accentUpper}-${gtkTheme}";
+      package = pkgs.catppuccin-gtk.override {
+        inherit (cfg) size tweaks;
+        accents = [cfg.accent];
```
Stonks3141, line 48:
```suggestion
        accents = [ cfg.accent ];
```

Stonks3141 (changes requested):
LGTM! I don't know if this is just what the formatter you're using does, but there is no whitespace around single-element arrays. If that is a formatter thing, ignore those suggestions and we'll figure that out when we decide on a formatter.

getchoo:
> I don't know if this is just what the formatter you're using does, but there is no whitespace around single-element arrays. If that is a formatter thing, ignore those suggestions and we'll figure that out when we decide on a formatter.

that was just my mistake lol

PR #8, opened 2023-04-14
add support for helix

getchoo:
this adds support for the [helix](https://github.com/catppuccin/helix) theme

### DIFF
--- a/modules/home-manager/default.nix
+++ b/modules/home-manager/default.nix
@@ -2,6 +2,7 @@
   imports = [
     ./bat.nix
     ./starship.nix
+    ./helix.nix
   ];
   options.catppuccin = {
     flavour = lib.mkOption {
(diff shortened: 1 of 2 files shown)


Stonks3141:
Helix bundles the Catppuccin themes by default, is there any reason to add them here?

getchoo:
i would like to say it was for customization...but honestly i misread the README and didn't catch that they were included (i don't use helix myself, it was just an easy theme that came to mind)

i still think fetching the repository could be a good idea though since it allows users to choose between the variants. if that's not good though, i'll just take out the theme declaration while leaving the settings

Stonks3141:
Oh yeah, I forgot about the no-italics variant. I'll merge this then.

Stonks3141 (approved):

PR #19, opened 2023-04-15
add support for kitty

getchoo:
this adds support for the [kitty](https://github.com/catppuccin/kitty) theme

i was also thinking of maybe making a utils file like my previous prs, but with general functions others might need to use (like with `mkUpper` in this case since it's also used by the gtk hm module). let me know what you think :)

### DIFF
--- a/modules/home-manager/kitty.nix
+++ b/modules/home-manager/kitty.nix
@@ -1,0 +1,23 @@
+{ config, lib, ... }:
+let cfg = config.programs.kitty.catppuccin;
+in {
+  options.programs.kitty.catppuccin = with lib; {
+    enable = mkEnableOption "Catppuccin theme";
+    flavour = mkOption {
+      type = types.enum [ "latte" "frappe" "macchiato" "mocha" ];
+      default = config.catppuccin.flavour;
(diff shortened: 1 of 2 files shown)


modules/home-manager/kitty.nix
```diff
@@ -0,0 +1,23 @@
...
+    };
+  };
+
+  config.programs.kitty = with lib;
+    let
+      # string -> string
+      # this capitalizes the first letter in a string
+      # it's used to set the theme name correctly here
+      mkUpper = word:
+        (toUpper (substring 0 1 word)) + (substring 1 (stringLength word) word);
```
Stonks3141, line 19: Can we move this into a `utils.nix` or something? Maybe also the `flavour` and `accent` types.

Stonks3141 (approved):
Thanks!

modules/home-manager/kitty.nix
getchoo: i was thinking of something like that as well
maybe with a `mkCatppuccinOption` function too in order to get rid of the boilerplate from the first part of the files all together?

PR #21, opened 2023-04-15
add support for tmux

getchoo:
this adds support for the [tmux](https://github.com/catppuccin/tmux) theme

### DIFF
--- a/modules/home-manager/tmux.nix
+++ b/modules/home-manager/tmux.nix
@@ -5,7 +5,7 @@ let
   plugin = with builtins;
     with pkgs;
     let rev = "4e48b09a76829edc7b55fbb15467cf0411f07931";
-    in pkgs.tmuxPlugins.mkTmuxPlugin {
+    in tmuxPlugins.mkTmuxPlugin {
       pluginName = "catppuccin";
       version = substring 0 7 rev;
       src = fetchFromGitHub {
--- a/modules/home-manager/tmux.nix
+++ b/modules/home-manager/tmux.nix
@@ -1,0 +1,33 @@
+{ config, pkgs, lib, ... }:
+let
+  cfg = config.programs.tmux.catppuccin;
+
+  plugin = with builtins;
+    with pkgs;
+    let rev = "4e48b09a76829edc7b55fbb15467cf0411f07931";
+    in pkgs.tmuxPlugins.mkTmuxPlugin {
+      pluginName = "catppuccin";
+      version = substring 0 7 rev;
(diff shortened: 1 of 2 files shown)


modules/home-manager/tmux.nix
```diff
@@ -0,0 +1,33 @@
+{ config, pkgs, lib, ... }:
+let
+  cfg = config.programs.tmux.catppuccin;
+
+  plugin = with builtins;
+    with pkgs;
+    let rev = "4e48b09a76829edc7b55fbb15467cf0411f07931";
+    in pkgs.tmuxPlugins.mkTmuxPlugin {
```
Stonks3141, line 8:
```suggestion
    in tmuxPlugins.mkTmuxPlugin {
```
`pkgs` is already in scope.

modules/home-manager/tmux.nix
```diff
@@ -0,0 +1,33 @@
...
+      pluginName = "catppuccin";
+      version = substring 0 7 rev;
```
Stonks3141, line 10: I assume this is a tmux thing, but can we use the full rev?

modules/home-manager/tmux.nix
```diff
@@ -0,0 +1,33 @@
...
+      default = config.catppuccin.flavour;
+      description = "Catppuccin flavour for tmux";
+    };
+  };
+
+  config.programs.tmux.plugins = with lib;
+    mkIf cfg.enable [{
+      inherit plugin;
+      extraConfig = "set -g @catppuccin_flavour '${cfg.flavour}'";
+    }];
```
Stonks3141, line 32:
```suggestion
  config.programs.tmux.plugins = with lib; mkIf cfg.enable [
    {
      inherit plugin;
      extraConfig = "set -g @catppuccin_flavour '${cfg.flavour}'";
    }
  ];
```
I think the new line for `cfg.enable` and `[{` just looks a bit gross. Ignore this if this code is after a formatter ran on it.

Stonks3141 (changes requested):

modules/home-manager/tmux.nix
getchoo: it isn't. most packages just slim down the commit hash or use another form of shorthand like a date, plus it's also what i usually do

Stonks3141: Oh, I didn't realize that was the version of the package, not the version to fetch. :)

Stonks3141 (approved):

PR #25, opened 2023-04-16
add util library

getchoo:
this adds (or opens up) a few helper expressions to the project:

- `mkUpper`

  - capitalizes the first letter of a string (sometimes needed for formatting)

- `mapModules`

  - this is used in order to share the extended library globally

- `types`

  - this is an attrset containing enums of flavours and accents, so we don't have to type so much anymore

- `mkBasicOpt`

  - this is mainly an internal function to generate options without too much boilerplate

- `mkFlavourOpt`

  - this creates a flavour option based on the psuedo-template we've been using

- `mkAccentOpt`

  - ditto, but with accents

- `mkCatppuccinOpt`
  - this is probably the most helpful since it basically removes the need to declare options for modules that only use `flavour` (the majority)
  
overall, i think these can all make contributing faster and easier, especially for regulars who become familiar with this library -- and unlike my previous pr, it won't change the workflow of those who prefer to just use standard declarations since it doesn't force any "magic." the only major complication this adds is the map at the end of `imports` arrays, but i think that's an okay tradeoff here given the benefits of these expressions

### DIFF
--- a/modules/home-manager/default.nix
+++ b/modules/home-manager/default.nix
@@ -1,16 +1,21 @@
-{ config, pkgs, lib, ... }: {
-  imports = [
-    ./alacritty.nix
-    ./bat.nix
-    ./bottom.nix
-    ./btop.nix
-    ./kitty.nix
-    ./starship.nix
-    ./helix.nix
-    ./gtk.nix
-    ./polybar.nix
-    ./tmux.nix
-  ];
+{ config, pkgs, lib, ... }: let
+ extendedLib = import ../lib/mkExtLib.nix lib;
+in {
+  imports = let
+    files = [
+      ./alacritty.nix
+      ./bat.nix
+      ./bottom.nix
+      ./btop.nix
+      ./kitty.nix
+      ./starship.nix
+      ./helix.nix
+      ./gtk.nix
+      ./polybar.nix
+      ./tmux.nix
+    ];
+  in extendedLib.ctp.mapModules config pkgs extendedLib files;
+
   options.catppuccin = {
     flavour = lib.mkOption {
       type = lib.types.enum [ "latte" "frappe" "macchiato" "mocha" ];
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -1,18 +1,10 @@
 { config, pkgs, lib, ... }:
 let cfg = config.gtk.catppuccin;
 in {
-  options.gtk.catppuccin = with lib; {
-    enable = mkEnableOption "Catppuccin theme";
-    flavour = mkOption {
-      type = types.enum [ "latte" "frappe" "macchiato" "mocha" ];
-      default = config.catppuccin.flavour;
-      description = "Catppuccin flavour for gtk";
-    };
-    accent = mkOption {
-      type = types.enum [ "blue" "flamingo" "green" "lavender" "maroon" "mauve" "peach" "pink" "red" "rosewater" "sapphire" "sky" "teal" "yellow" ];
-      default = config.catppuccin.accent;
-      description = "Catppuccin accents for gtk";
-    };
+  options.gtk.catppuccin =
+    lib.ctp.mkCatppuccinOpt "gtk" config // (
+    with lib; {
+    accent = ctp.mkAccentOpt "gtk" config;
     size = mkOption {
       type = types.enum [ "standard" "compact" ];
       default = "standard";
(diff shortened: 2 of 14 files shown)


modules/home-manager/gtk.nix
```diff
@@ -1,18 +1,10 @@
...
-    };
-    accent = mkOption {
-      type = types.enum [ "blue" "flamingo" "green" "lavender" "maroon" "mauve" "peach" "pink" "red" "rosewater" "sapphire" "sky" "teal" "yellow" ];
-      default = config.catppuccin.accent;
-      description = "Catppuccin accents for gtk";
-    };
+  options.gtk.catppuccin =
+    lib.ctp.mkCatppuccinOpt "gtk" config // (
+    with lib; {
+    accent = ctp.mkAccentOpt "gtk" config;
```
Stonks3141, line 7:
```suggestion
  options.gtk.catppuccin = with lib;
    ctp.mkCatppuccinOpt "gtk" config // {
    accent = ctp.mkAccentOpt "gtk" config;
```

modules/home-manager/gtk.nix
```diff
@@ -23,19 +15,14 @@ in {
       default = [ "normal" ];
       description = "Catppuccin tweaks for gtk";
     };
-  };
+  });
```
Stonks3141, line 18:
```suggestion
  };
```

modules/home-manager/helix.nix
```diff
@@ -1,15 +1,10 @@
...
 in {
-  options.programs.helix.catppuccin = with lib; {
-    enable = mkEnableOption "Catppuccin theme";
-    flavour = mkOption {
-      type = types.enum [ "latte" "frappe" "macchiato" "mocha" ];
-      default = config.catppuccin.flavour;
-      description = "Catppuccin flavour for Helix";
+  options.programs.helix.catppuccin =
+    lib.ctp.mkCatppuccinOpt "helix" config // {
+      useItalics = lib.mkEnableOption "Italics in Catppuccin theme for Helix";
```
Stonks3141, line 6:
```suggestion
  options.programs.helix.catppuccin = with lib;
    ctp.mkCatppuccinOpt "helix" config // {
      useItalics = mkEnableOption "Italics in Catppuccin theme for Helix";
```

modules/home-manager/kitty.nix
```diff
@@ -1,23 +1,11 @@
...
-    let
-      # string -> string
-      # this capitalizes the first letter in a string
-      # it's used to set the theme name correctly here
-      mkUpper = word:
-        (toUpper (substring 0 1 word)) + (substring 1 (stringLength word) word);
-
-      flavourUpper = mkUpper cfg.flavour;
+    with ctp;
+    let flavourUpper = mkUpper cfg.flavour;
```
Stonks3141, line 9:
```suggestion
    let flavourUpper = ctp.mkUpper cfg.flavour;
```

modules/lib/default.nix
```diff
@@ -0,0 +1,86 @@
+lib:
+with builtins;
+with lib; rec {
+  # string -> string
+  # this capitalizes the first letter in a string,
+  # which is sometimes needed in order to format
+  # the names of themes correctly
+  mkUpper = word:
+    (toUpper (substring 0 1 word)) + (substring 1 (stringLength word) word);
```
Stonks3141, line 9:
```suggestion
  mkUpper = str:
    (toUpper (substring 0 1 str)) + (substring 1 (stringLength str) str);
```
`word` doesn't make much sense now that this is a library function.

modules/nixos/default.nix
```diff
@@ -1,4 +1,13 @@
...
+  imports = let
+    files = [
+    ];
+  in
+    extendedLib.ctp.mapModules config pkgs extendedLib files;
+
+
   options.catppuccin = {
     flavour = lib.mkOption {
       type = lib.types.enum [ "latte" "frappe" "macchiato" "mocha" ];
```
Stonks3141, line 13:
```suggestion
      type = extendedLib.ctp.types.flavourOption;
```

modules/home-manager/default.nix
```diff
@@ -1,16 +1,21 @@
...
+      ./helix.nix
+      ./gtk.nix
+      ./polybar.nix
+      ./tmux.nix
+    ];
+  in extendedLib.ctp.mapModules config pkgs extendedLib files;
+
   options.catppuccin = {
     flavour = lib.mkOption {
       type = lib.types.enum [ "latte" "frappe" "macchiato" "mocha" ];
```
Stonks3141, line 21:
```suggestion
      type = extendedLib.ctp.types.flavourOption;
```
Do this for the accent option too, Github won't let me suggest changes to lines that aren't in the diff.

Stonks3141 (changes requested):
Thanks! Just a few minor style issues.

Stonks3141 (dismissed):

PR #27, opened 2023-04-18
add support for neovim

getchoo:
this adds support for the [neovim](https://github.com/catppuccin/nvim) theme


Stonks3141 (approved):
Thanks! I'll just have to assume this works ;)

PR #32, opened 2023-06-02
vendor nixpkgs, cleanup some stuff

getchoo:
as the title says, this vendors nixpkgs, fixing possible issues for those using outdated revisions. with this change, the formatter can also now be set in a flake attribute and `nix fmt` can be used.

i also fixed a bug in the module for bat, which blocked hm activation packages from building due to the command we set now being a default :)

### DIFF
--- a/modules/home-manager/bat.nix
+++ b/modules/home-manager/bat.nix
@@ -5,8 +5,6 @@ let cfg = config.programs.bat.catppuccin; in
     lib.ctp.mkCatppuccinOpt "bat" config;
 
   config = {
-    home.activation.batCache = "${pkgs.bat}/bin/bat cache --build";
-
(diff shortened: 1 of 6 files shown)


modules/home-manager/bat.nix
```diff
@@ -5,8 +5,6 @@ let cfg = config.programs.bat.catppuccin; in
     lib.ctp.mkCatppuccinOpt "bat" config;
 
   config = {
-    home.activation.batCache = "${pkgs.bat}/bin/bat cache --build";
```
Stonks3141, line 8: Why is this removed?

Stonks3141: Never mind, I missed the commit message.

Stonks3141 (approved):
Thanks!

PR #44, opened 2023-10-16
feat(modules/home-manager): add glamour

ryanccn:
Adds support for [glamour](https://github.com/charmbracelet/glamour), a Markdown rendering library used in a variety of Go CLIs (including `gh` and `glow`).


ryanccn:
🔥

PR #55, opened 2023-11-03
feat(home-manager): init wofi module

Anomalocaridid:
Adds support for wofi.

In case it matters, the Catppuccin GitHub organization declined taking the theme's repo because of how wofi is technically unmaintained: https://github.com/catppuccin/catppuccin/issues/2165.

The theme has a specific font hardcoded, so I took the liberty of making a font option for it for people that wish to override it.

### DIFF
--- a/_sources/generated.json
+++ b/_sources/generated.json
@@ -358,5 +358,25 @@
             "type": "github"
         },
         "version": "47e33044b4b47b1c1faca1e42508fc92be12131a"
+    },
+    "wofi": {
+        "cargoLocks": null,
+        "date": "2023-09-16",
+        "extract": null,
+        "name": "wofi",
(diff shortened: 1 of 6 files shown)


getchoo:
> In case it matters, the Catppuccin GitHub organization declined taking the theme's repo because of how wofi is technically unmaintained: https://github.com/catppuccin/catppuccin/issues/2165.

i'm going to have to agree with hammy here, as well as state that i don't think i'll be accepting unofficial ports into this repository anytime soon. i'll be adding this to the contribution guidelines now, as i would rather not have anyone spending time on these when they won't be accepted. i'm sorry i didn't have it there in the first place :(

Anomalocaridid:
Fair enough. Thanks anyways.

PR #66, opened 2023-12-20
fix(home-manager/bat): wrong themeName

birdmanmandbir:
fix warning of bat:
>[bat warning]: Unknown theme 'Catppuccin-mocha', using default.

solve #67 

## my test
no warning with this pr
image


getchoo:
see my [comment](https://github.com/Stonks3141/ctp-nix/issues/67#issuecomment-1885741376) in #67, this might not be needed

PR #76, opened 2024-01-30
fix: apply lazygit theme

zspher:
use `themes-mergable` folder to apply lazygit theme as this has `gui` field

see https://github.com/catppuccin/lazygit/issues/13


getchoo:
thanks!

PR #80, opened 2024-02-08
feat(home-manager/hyprland): source theme + accent

Liassica:
Since Home Manager defbb9c, Hyprland on Home Manager now sources files before most other configuration, meaning ctp-nix can now source a theme colors file instead of inserting them into the main configuration. In addition, accent and accentAlpha color variables have been added in a second sourced file.

### DIFF
--- a/modules/home-manager/hyprland.nix
+++ b/modules/home-manager/hyprland.nix
@@ -1,19 +1,25 @@
-{ config
-, lib
-, sources
-, ...
+{
+  config,
+  lib,
+  sources,
+  ...
 }:
 let
   cfg = config.wayland.windowManager.hyprland.catppuccin;
   enable = cfg.enable && config.wayland.windowManager.hyprland.enable;
 in
 {
-  options.wayland.windowManager.hyprland.catppuccin =
-    lib.ctp.mkCatppuccinOpt "hyprland";
+  options.wayland.windowManager.hyprland.catppuccin = lib.ctp.mkCatppuccinOpt "hyprland" // {
+    accent = lib.ctp.mkAccentOpt "hyprland";
+  };
 
-  # Because of how nix merges options and hyprland interprets options, sourcing
-  # the file does not work. instead, parse the theme and put it in settings so
-  # the variables appear early enough in the file to be applied properly.
-  config.wayland.windowManager.hyprland.settings = lib.mkIf enable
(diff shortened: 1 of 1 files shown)


Anomalocaridid:
If files are sourced first, would it be possible to implement an accent option that sets Hyprland variables named something like `$accent` and `$accentAlpha` to the corresponding variables for your accent color?

For example, if your accent is red, would the below lines work if the module added them to your Hyprland config?
```hypr
$accent=$red
$accentAlpha=$redAlpha
```

Liassica:
Added that. Currently it's implemented as a second sourced file, as variables are one of the only configuration that gets placed before `source` entries, so they can't just be added with `hyprland.settings`.

modules/home-manager/hyprland.nix
```diff
@@ -1,19 +1,18 @@
...
-  # the variables appear early enough in the file to be applied properly.
-  config.wayland.windowManager.hyprland.settings = lib.mkIf enable
-    (lib.ctp.fromINI (sources.hyprland + /themes/${cfg.flavour}.conf));
+  config.wayland.windowManager.hyprland.settings.source = [
+    (lib.mkIf enable (sources.hyprland + /themes/${cfg.flavour}.conf))
+    (builtins.toFile "hyprland-${cfg.accent}-accent.conf" ''
+      $accent=''$${cfg.accent}
+      $accentAlpha=''$${cfg.accent}Alpha
+    '')
+  ];
```
getchoo, line 17:
```suggestion
  config.wayland.windowManager.hyprland.settings = lib.mkIf enable {
    source = [
      "${sources.hyprland}/themes/${cfg.flavour}.conf"
      (builtins.toFile "hyprland-${cfg.accent}-accent.conf" ''
        $accent=''$${cfg.accent}
        $accentAlpha=''$${cfg.accent}Alpha
      '')
    ];
  };
```

getchoo (changes requested):

getchoo:
sorry for the wait! it seems this branch has fallen a bit behind and is still stuck with a bug that's been fixed; could this be rebased?

Liassica:
Done! I also updated the commit description a bit since this is an official Catppuccin repository now.

getchoo (approved):
so sorry for the long wait. thanks so much!

PR #82, opened 2024-02-16
feat(home-manager): init delta module

Anomalocaridid:
Adds support for delta.

### DIFF
--- a/test.nix
+++ b/test.nix
@@ -57,6 +57,8 @@ in
         bottom = ctpEnable;
         btop = ctpEnable;
         fish = ctpEnable;
+        git.enable = true; # Required for delta
+        git.delta = ctpEnable;
         glamour.catppuccin.enable = true;
         helix = ctpEnable;
         home-manager.enable = false;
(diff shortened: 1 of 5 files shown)


test.nix
```diff
@@ -57,6 +57,7 @@ in
         bottom = ctpEnable;
         btop = ctpEnable;
         fish = ctpEnable;
+        git.delta = ctpEnable;
```
getchoo, line 60: git also needs to be enabled for this to be applied - otherwise this works great. thanks!

getchoo (changes requested):

test.nix
Anomalocaridid: Git is now enabled in test.nix

Anomalocaridid:
I also had to update the delta source and slightly change how the theme is applied due to some upstream changes

isabelroses (approved):

getchoo (approved):

getchoo:
thanks!

PR #93, opened 2024-03-23
feat(home-manager): init fzf module

Anomalocaridid:
Adds support for fzf.

I took the same approach that the console module uses and used the palette to manually add the colors from [catppuccin/fzf](https://github.com/catppuccin/fzf).

### DIFF
--- a/modules/home-manager/fzf.nix
+++ b/modules/home-manager/fzf.nix
@@ -1,0 +1,35 @@
+{ config
+, lib
+, sources
+, ...
+}:
+let
+  cfg = config.programs.fzf.catppuccin;
+  enable = cfg.enable && config.programs.fzf.enable;
+  palette = (lib.importJSON "${sources.palette}/palette.json").${cfg.flavour}.colors;
+in
+{
+  options.programs.fzf.catppuccin =
+    lib.ctp.mkCatppuccinOpt "fzf";
+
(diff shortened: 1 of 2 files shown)


modules/home-manager/fzf.nix
```diff
@@ -0,0 +1,34 @@
...
+    lib.ctp.mkCatppuccinOpt "fzf";
+
+  config.programs.fzf.colors = lib.mkIf enable
+    # Manually populate with colors from catppuccin/fzf
+    (lib.attrsets.mapAttrs (_: color: palette.${color}.hex)
+      {
+        "bg+" = "surface0";
+        bg = "base";
+        spinner = "rosewater";
+        hl = "red";
```
isabelroses, line 22: nit: group hl with hl+ and the same for fg and fg+ like you did for bg and bg+

Anomalocaridid: I can do that, but first I should mention I put the options in that specific order to match the (seemingly arbitrary) order of the options in catppuccin/fzf to make it easier to compare.

isabelroses: Ahhh, that makes total sense.

getchoo:
this needs a rebase

Anomalocaridid:
Rebased

getchoo (approved):
thanks!

PR #97, opened 2024-04-02
feat(modules): expose palette in config.lib

### LINKED ISSUE #96: expose palette to lib
to allow usage for applications/modules that don't yet have a port

Anomalocaridid:
Exposes palettes to `config.lib` in both NixOS and Home Manager modules. Closes #96.

I essentially took `palette.json` from [catppuccin/palette](https://github.com/catppuccin/palette) and converted it into an attrset, `config.lib.catppuccin.palettes`. I also made it so the palette corresponding to the user's set flavor would be exposed by itself as `config.lib.catppuccin.palette`. In the home manager module, the latter also has an extra attribute that corresponds to the user's set accent color for convenience.

### DIFF
--- a/modules/home-manager/globals.nix
+++ b/modules/home-manager/globals.nix
@@ -1,4 +1,17 @@
-{ lib, ... }: {
+{ config, lib, sources, ... }: {
+  config.lib.catppuccin =
+    let
+      cfg = config.catppuccin;
+      palettes = lib.importJSON "${sources.palette}/palette.json";
+      palette = palettes.${cfg.flavour};
+    in
+    {
+      inherit palettes;
+      palette = lib.attrsets.recursiveUpdate palette {
+        colors.accent = palette.colors.${cfg.accent};
+      };
+    };
+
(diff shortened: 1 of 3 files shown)


getchoo:
sorry for the wait, but i wanted to explore some alternatives before really getting into this

i don't think exposing `config.lib` is the best way to go here, or only exposing the palette itself. most of my thoughts are still summed in https://github.com/catppuccin/nix/issues/96#issuecomment-2041403103, but to get more technical: `sources` should probably be a global option that's a name value pair of the source name and it's path in the store. this would allow for any source to be reused or changed fairly easily, and still be similar to how `sources` is an argument for our modules now. i have my own branch with this idea that i will be PRing soon, but it also will (probably) require a switch to `npins` from `nvfetcher`, which is something i'd like to handle myself

thanks for your work here though, i genuinely appreciate it :)

Anomalocaridid:
No worries. Honestly, that does seem like a much better and more general approach.

PR #98, opened 2024-04-08
feat(home-manager): init gitui module

### LINKED ISSUE #13: Port request: gitui
Port: https://github.com/catppuccin/gitui
Example: https://git.sr.ht/~stonks3141/nixos-config/tree/main/item/modules/applications/gitui.nix

sophieforrest:
Adds support for gitui. Resolves #13.

### DIFF
--- a/modules/home-manager/gitui.nix
+++ b/modules/home-manager/gitui.nix
@@ -1,0 +1,20 @@
+{ config
+, lib
+, sources
+, ...
+}:
+let
+  cfg = config.programs.gitui.catppuccin;
+  enable = cfg.enable && config.programs.gitui.enable;
+in
+{
+  options.programs.gitui.catppuccin =
+    lib.ctp.mkCatppuccinOpt "gitui";
+
+  config = lib.mkIf enable {
+    programs.gitui.theme = builtins.path {
+      name = "${cfg.flavour}.ron";
+      path = "${sources.gitui}/theme/${cfg.flavour}.ron";
+    };
+  };
+}
--- a/_sources/generated.json
+++ b/_sources/generated.json
@@ -119,6 +119,26 @@
         },
         "version": "0ce27b518e8ead555dec34dd8be3df5bd75cff8e"
     },
+    "gitui": {
+        "cargoLocks": null,
+        "date": "2023-11-13",
+        "extract": null,
+        "name": "gitui",
+        "passthru": null,
+        "pinned": false,
+        "src": {
+            "deepClone": false,
+            "fetchSubmodules": false,
+            "leaveDotGit": false,
+            "name": null,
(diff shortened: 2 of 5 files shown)


modules/home-manager/gitui.nix
```diff
@@ -0,0 +1,17 @@
...
+let
+  cfg = config.programs.gitui.catppuccin;
+  enable = cfg.enable && config.programs.gitui.enable;
+in
+{
+  options.programs.gitui.catppuccin =
+    lib.ctp.mkCatppuccinOpt "gitui";
+
+  config = lib.mkIf enable {
+    programs.gitui.theme = builtins.readFile "${sources.gitui}/theme/${cfg.flavour}.ron";
```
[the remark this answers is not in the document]
sophieforrest: Hey, I'm not exactly sure how I would go about doing this. The obvious approach to me would be to do something to the effect of `/. + ${sources.gitui}/theme/${cfg.flavour}.ron` to get a path, however this isn't permitted by Nix as you cannot append store paths to other paths.

Wondering if I might be missing something here? Thanks for your input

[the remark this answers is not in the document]
isabelroses: Seem like you should be fine comparing to other themes https://github.com/catppuccin/nix/blob/4043beb6c792a40e9c513aa261f4ba10601a4457/modules/home-manager/polybar.nix#L15

getchoo, line 15:
```suggestion
    programs.gitui.theme = "${sources.gitui}/theme/${cfg.flavour}.ron";
```
isabel was right, this can just be a path

getchoo (changes requested):

modules/home-manager/gitui.nix
sophieforrest: Hey, just double checked this in case I didn't try this.

From my testing this seems to just write the store path to `theme.ron` instead of linking to it. I suspect this is because the two checks home-manager does to detect if it is a file (`builtins.isPath` and `lib.isStorePath`) both return `false` when compared against the path. The tricky one is `builtins.isPath`, because it seems to only return `true` if the provided argument is of type `path`, but with the interpolation it ends up being a string with context.

Would appreciate any ideas you have on how to solve this

getchoo: ah, sorry about that. `builtins.path` should be able to help here
```suggestion
    programs.gitui.theme = builtins.path {
      name = "${cfg.flavour}.ron";
      path = "${sources.gitui}/theme/${cfg.flavour}.ron";
    };
```

sophieforrest: Works perfectly! Thank you!

getchoo:
thanks!

PR #104, opened 2024-04-09
feat(home-manager): add support for dunst

### LINKED ISSUE #99: Add Dunst Support
I don't think this should be too hard. You already have Dunst themes here:
https://github.com/catppuccin/dunst/tree/main

For Home-Manager integration, you use one of these two options:
https://nix-community.github.io/home-manager/options.xhtml#opt-services.dunst.configFile
https://nix-community.github.io/home-manager/options.xhtml#opt-services.dunst.settings

henrisota:
Adds support for dunst. Resolves #99

### DIFF
--- a/modules/home-manager/dunst.nix
+++ b/modules/home-manager/dunst.nix
@@ -1,0 +1,18 @@
+{ config
+, lib
+, sources
+, ...
+}:
+let
+  inherit (lib) ctp;
+  cfg = config.services.dunst.catppuccin;
+  enable = cfg.enable && config.services.dunst.enable;
+in
+{
+  options.services.dunst.catppuccin =
+    lib.ctp.mkCatppuccinOpt "dunst";
+
+  config.services.dunst = lib.mkIf enable {
+    settings = lib.ctp.fromINI (sources.dunst + /themes/${cfg.flavour}.conf);
+  };
+}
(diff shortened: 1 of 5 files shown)


modules/home-manager/dunst.nix
```diff
@@ -0,0 +1,18 @@
...
+  cfg = config.services.dunst.catppuccin;
+  enable = cfg.enable && config.services.dunst.enable;
+in
+{
+  options.services.dunst.catppuccin =
+    lib.ctp.mkCatppuccinOpt "dunst";
+
+  config.services.dunst.settings =
+    lib.mkIf enable
+      (lib.ctp.fromINI (sources.dunst + /themes/${cfg.flavour}.conf));
```
getchoo, line 17:
```suggestion
  config.services.dunst = lib.mkIf enable {
    settings = lib.ctp.fromINI (sources.dunst + /themes/${cfg.flavour}.conf);
  };
```

getchoo (changes requested):

modules/home-manager/dunst.nix
henrisota: Commited with the change.

getchoo (approved):
thanks :)

PR #119, opened 2024-04-15
fix(home-manager): add file for yazi syntax highlighting

### LINKED ISSUE #111: yazi: theme installs but no synatx highlighting
The yazi theme decalres the synax highligting to `~/.config/yazi/theme.tmTheme`
see https://github.com/catppuccin/yazi/blob/0846aed69b2a62d29c98e100af0cf55ca729723d/themes/mocha.toml#L32

But the flake does not provide that file, this could be solved by useing the bat themes' .tmTheme file.

michaelBelsanti:
Fixes #111

### DIFF
--- a/modules/home-manager/yazi.nix
+++ b/modules/home-manager/yazi.nix
@@ -13,6 +13,11 @@ in
     ctp.mkCatppuccinOpt "yazi";
 
   config = lib.mkIf enable {
+    assertions = [
+      (lib.ctp.assertXdgEnabled "yazi")
+    ];
+
     programs.yazi.theme = lib.importTOML "${sources.yazi}/themes/${cfg.flavour}.toml";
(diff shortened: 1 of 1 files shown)


isabelroses (approved):

getchoo (approved):
thanks!

modules/home-manager/yazi.nix
```diff
@@ -14,5 +14,6 @@ in
 
   config = lib.mkIf enable {
```
getchoo, line 15:
```suggestion
  config = lib.mkIf enable {
    assertions = [
      (lib.ctp.assertXdgEnabled "yazi")
    ];
```
oops, might have been a bit too eager 😆 
an assertion will need to be used here to make sure `xdg` is enabled

getchoo (changes requested):
oops, was a little bit too quick there.

modules/home-manager/yazi.nix
michaelBelsanti: good catch, forgot that you could disable xdg

getchoo (approved):
thanks!

PR #121, opened 2024-04-15
feat(home-manager): add support for cava

henrisota:
Adds support for cava.

I went ahead with duplicating the code for raw parsing the ini file (by using the `-r` parameter for `jc`) as `cava` requires the quotation marks for recognizing hexadecimal color strings.


getchoo (approved):
thanks!

Issue #122, opened 2024-04-17
Rofi support forces a default option for font that might not work with all users


IQubic:
The rofi themes provided by a recent commit seem to just use the default theme configs from https://github.com/catppuccin/rofi/tree/main/basic/.local/share/rofi/themes

The main issue with this is these 4 themes all have the same line 12: `font: "JetBrainsMono Nerd Font 14";`  This font isn't installed on my machine, and the options provided here don't give me any ways to change this. I'd prefer this to either install the required fonts for me, or give me a way to change the font myself.

aloop:
Not sure if it is compatible with the way this module sets up the rofi theme, but in my configuration I use the theme and override the defaults like so:

```nix
{
  programs.rofi = {
    # ...
    theme = {
      "@theme" = theme.package + /rofi/catppuccin-${theme.variant}.rasi;

      "*" = {
        font = "monospace 14";
      };
    };
  };
}
```

Where `theme.package` is pkgs.catppuccin.

I'm sure the module could be adapted to support this.

PR #123, opened 2024-04-17
fix(home-manager): allow overriding styles for the rofi theme

aloop:
Fix for #122.

Alters `programs.rofi.theme` to use an attrset instead of a path directly, so that users can add overriding styles to `programs.rofi.theme`.

A minimal example of how to change the font used in the theme would look like the following:

```nix
{
  programs.rofi = {
    enable = true;

    catppuccin.enable = true;

    theme."*".font = "monospace 14";
}
```

I've tested the changes with my NixOS/home-manager configuration, and it seems to work fine.


getchoo (approved):
tysm!

Issue #115, opened 2024-04-12
Add global `enableAll` setting


becknik:
Hi,

I'd love to just add this flake to my inputs and use a global toggle to enable the theming on all programs I've installed.
This would've the benefit of not having to check on the progress of the program options now and then and to automatically get satisfied when there is a new program supported :)

What do you think?

PS: Thanks for your work!

drupol:
Temporary solution: https://discourse.nixos.org/t/catppuccin-nix-the-soothing-pastel-theme-but-for-nix/42915/13?u=drupol

PR #124, opened 2024-04-17
feat: add `enable` option globally

drupol:
This PR:

- Add a new `enable` option to enable `catppuccin` everywhere it can
- Should fix https://github.com/catppuccin/nix/issues/115

I won't be able to test this until late tonight, help needed!

### DIFF
--- a/modules/home-manager/globals.nix
+++ b/modules/home-manager/globals.nix
@@ -1,5 +1,7 @@
 { lib, ... }: {
   options.catppuccin = {
+    enable = lib.mkEnableOption "Catppuccin globally";
+
     flavour = lib.mkOption {
       type = lib.ctp.types.flavourOption;
       default = "latte";
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -105,7 +105,9 @@ in
   # be the name of the module, followed by the local config
   # attrset
   mkCatppuccinOpt = name: {
-    enable = lib.mkEnableOption "Catppuccin theme";
+    enable = lib.mkEnableOption "Catppuccin theme" // {
+      default = config.catppuccin.enable;
+    };
     flavour = mkFlavourOpt name;
   };
 
--- a/modules/nixos/globals.nix
+++ b/modules/nixos/globals.nix
@@ -1,5 +1,7 @@
 { lib, ... }: {
   options.catppuccin = {
+    enable = lib.mkEnableOption "Catppuccin globally";
+
     flavour = lib.mkOption {
       type = lib.ctp.types.flavourOption;
       default = "latte";
--- a/dev/flake.lock
+++ b/dev/flake.lock
@@ -1,20 +1,5 @@
 {
   "nodes": {
-    "get-flake": {
-      "locked": {
-        "lastModified": 1694475786,
-        "narHash": "sha256-s5wDmPooMUNIAAsxxCMMh9g68AueGg63DYk2hVZJbc8=",
-        "owner": "ursi",
-        "repo": "get-flake",
-        "rev": "ac54750e3b95dab6ec0726d77f440efe6045bec1",
-        "type": "github"
-      },
-      "original": {
-        "owner": "ursi",
-        "repo": "get-flake",
-        "type": "github"
-      }
-    },
     "home-manager": {
       "inputs": {
         "nixpkgs": [
@@ -52,7 +37,6 @@
     },
     "root": {
       "inputs": {
-        "get-flake": "get-flake",
         "home-manager": "home-manager",
         "nixpkgs": "nixpkgs"
       }
--- a/dev/flake.nix
+++ b/dev/flake.nix
@@ -8,11 +8,9 @@
       url = "github:nix-community/home-manager";
       inputs.nixpkgs.follows = "nixpkgs";
     };
-
-    get-flake.url = "github:ursi/get-flake";
   };
 
-  outputs = { self, nixpkgs, get-flake, ... }@inputs:
+  outputs = { self, nixpkgs, ... }@inputs:
     let
       systems = [
         "x86_64-linux"
@@ -24,11 +22,10 @@
       inherit (nixpkgs) lib;
 
       forAllSystems = fn: lib.genAttrs systems (s: fn nixpkgs.legacyPackages.${s});
-      ctp = get-flake ../.;
     in
     {
       checks = forAllSystems (pkgs: lib.optionalAttrs pkgs.stdenv.isLinux {
-        module-vm-test = pkgs.nixosTest (import ../test.nix { inherit ctp inputs; });
+        module-vm-test = pkgs.nixosTest (import ../test.nix inputs);
       });
 
       formatter = forAllSystems (pkgs: pkgs.nixpkgs-fmt);
@@ -57,8 +54,8 @@
             '';
         in
         {
-          nixos-doc = mkDoc "nixos" ctp.nixosModules.catppuccin;
-          home-manager-doc = mkDoc "home-manager" ctp.homeManagerModules.catppuccin;
+          nixos-doc = mkDoc "nixos" ../modules/nixos;
+          home-manager-doc = mkDoc "home-manager" ../modules/home-manager;
 
           default = self.packages.${pkgs.system}.home-manager-doc;
         });
--- a/test.nix
+++ b/test.nix
@@ -1,36 +1,35 @@
-{ ctp
-, inputs
-, ...
-}:
+inputs:
 let
   common = {
-    catppuccin.flavour = "mocha";
-    users.users.test = {
-      isNormalUser = true;
-      home = "/home/test";
+    catppuccin = {
+      enable = true;
+      flavour = "mocha";
     };
   };
 
-  ctpEnable = {
-    enable = true;
-    catppuccin.enable = true;
-  };
+  # shorthand enable
+  enable = { enable = true; };
 in
 {
   name = "module-test";
 
   nodes.machine = { lib, ... }: {
     imports = [
-      ctp.nixosModules.catppuccin
       inputs.home-manager.nixosModules.default
+      ./modules/nixos
       common
     ];
 
-    boot.loader.grub = ctpEnable;
+    boot.loader.grub = enable;
 
-    console = ctpEnable;
+    console = enable;
 
-    programs.dconf.enable = true; # required for gtk
+    programs.dconf = enable; # required for gtk
+
+    users.users.test = {
+      isNormalUser = true;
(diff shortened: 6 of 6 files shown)


isabelroses (commented):
Why not put it in [globals.nix](https://github.com/catppuccin/nix/blob/372decb578f3187d7aad837885ed98e73e8abd60/modules/home-manager/globals.nix).

And I don't belive you need the `flavour = lib.mkDefault cfg.flavour;` because of the aforementioned globals.

drupol:
doing it

drupol:
Just tested on my own local config with:

```nix
{ inputs, ... }:

{
  imports = [
    inputs.catppuccin.homeManagerModules.catppuccin
  ];

  catppuccin = {
    enable = true;
    flavour = "latte";
  };
}
```

And that works :)

modules/home-manager/default.nix
```diff
@@ -1,4 +1,23 @@
...
+let
+  cfg = config.programs.catppuccin-global;
+in
 {
   imports = import ../lib/mkImports.nix args ./.;
+
+  options.programs.catppuccin = lib.ctp.mkCatppuccinOpt "catppuccin global";
+
+  config.programs = lib.mkIf cfg.enable (
+    lib.foldlAttrs
```
drupol, line 16: I wish I could have used `builtins.mapAttrs`, but I couldn't get it done with it, I don't know why yet. Do you have a clue?

netlooker:
Many thanks @drupol as I was pulling my hair out to make it work 🙏

drupol:
@isabelroses Anything else left to do in here?

isabelroses:
> @isabelroses Anything else left to do in here?

Not in my opinion but ultimately its up to @getchoo.

drupol:
I just fixed it.

drupol:
I'm using `nixfmt-rfc-style` but going to fix the PR with your style just now.

modules/home-manager/globals.nix
```diff
@@ -11,5 +20,22 @@
       default = "teal";
       description = "Global Catppuccin accent";
     };
+
+    enable = lib.mkEnableOption "catppuccin globally";
```
getchoo, line 24:
```suggestion
    enable = lib.mkEnableOption "Catppuccin globally";
```

modules/home-manager/globals.nix
```diff
@@ -11,5 +20,22 @@
...
+        programs: name: value:
+          programs
+          // (lib.optionalAttrs (lib.hasAttr "catppuccin" value) {
+            ${name}.catppuccin = {
+              enable = lib.mkDefault cfg.enable;
+            };
+          })
+      )
+      { }
+      options.programs
```
getchoo, line 39: i don't think this is the best way to go about this. the biggest issue here is the assumption that all the options for each port are in `options.programs` when this isn't the case - mako and dunst's options are under `services`, for example. i would also considered it a bit of a forced approach

imo the best solution would be editing `lib.ctp.mkCatppuccinOpt` to make the `enable` and `flavour` option shared by all modules default to `config.catppuccin.enable`/`config.catppuccin.flavour`

modules/home-manager/globals.nix
getchoo, line 1: there should be an equivalent in the nixos model as well

getchoo (changes requested):
thanks for getting the ball rolling here btw :)

modules/home-manager/globals.nix
drupol: Fair enough... I'll see what I can do.

modules/home-manager/globals.nix
drupol: I've got no clue, sorry :S

modules/home-manager/globals.nix
isabelroses: Also just realised this does not account for `delta` which is under `programs.git.delta`

drupol: Would this be a better option ?

drupol:
I updated the PR, WDYT ?

pluiedev (changes requested):

modules/home-manager/alacritty.nix
```diff
@@ -6,7 +6,7 @@
 let
   inherit (lib) ctp;
   cfg = config.programs.alacritty.catppuccin;
-  enable = cfg.enable && config.programs.alacritty.enable;
+  enable = (cfg.enable && config.programs.alacritty.enable) || config.catppuccin.enable) || config.catppuccin.enable;
```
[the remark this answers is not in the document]
drupol: I would like to have some guidance on this.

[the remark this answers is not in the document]
isabelroses: Its a bit rushed buy maybe something like this
`isCatppuccinEnabled = conf: prog: conf.catppuccin.enable && conf.programs.${prog}.enable && conf.programs.${prog}.catppuccin.enable;`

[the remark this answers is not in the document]
drupol: I'll try something.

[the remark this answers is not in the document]
pluiedev: I think something like this would work:
```nix
# lib/default.nix
{config, ...}: {
	# ...
	isEnabled = cfg: config.catppuccin.enable || cfg.enable && cfg.catppuccin.enable; 
}

# modules/home-manager/alacritty.nix
{config, lib, ...}:
let
	inherit (lib) ctp;
	enable = ctp.isEnabled config.programs.alacritty;
in { ... }
```

But then, you're leaving out the option for users to disable theming as an override. IMO the more reasonable thing to do is to just make the default for each enable option to be `config.catppuccin.enable`, [which I believe is already the case](https://github.com/catppuccin/nix/blob/75663896d0c16cd59d567f21f091b1c9338d7118/modules/lib/default.nix#L16)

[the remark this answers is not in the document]
isabelroses: good catch, totally forgot to add the all enabled case.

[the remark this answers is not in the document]
drupol: I just pushed something, let me know what you think.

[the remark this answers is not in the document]
drupol: Very cool idea, I just pushed everything again ;)

modules/home-manager/globals.nix
```diff
@@ -11,5 +11,7 @@
       default = "teal";
       description = "Global Catppuccin accent";
     };
+
+    enable = lib.mkEnableOption "Catppuccin globally";
```
isabelroses, line 15: you should copy this to here too https://github.com/catppuccin/nix/blob/b87e5ee889aaeadc110fa8cf0c48e4115b39137c/modules/nixos/globals.nix#L2

drupol: fixed

isabelroses:
Probably should rename PR to `feat: enable all option`

docs/home-manager-options.md
getchoo, line 1: this should be dropped from the PR as we update this on each push, not in forks

modules/home-manager/globals.nix
getchoo: nit: i'd prefer this to go above the other options

modules/lib/default.nix
```diff
@@ -105,7 +105,9 @@ in
   # be the name of the module, followed by the local config
   # attrset
   mkCatppuccinOpt = name: {
-    enable = lib.mkEnableOption "Catppuccin theme";
+    enable = (lib.mkEnableOption "Catppuccin theme") // {
```
getchoo, line 108:
```suggestion
    enable = lib.mkEnableOption "Catppuccin theme" // {
```

modules/nixos/globals.nix
```diff
@@ -1,9 +1,13 @@
-{ lib, ... }: {
+{ lib, config, ... }: {
```
getchoo, line 1:
```suggestion
{ lib, ... }: {
```

modules/nixos/globals.nix
```diff
@@ -1,9 +1,13 @@
...
   options.catppuccin = {
     flavour = lib.mkOption {
       type = lib.ctp.types.flavourOption;
       default = "latte";
       description = "Global Catppuccin flavour";
     };
+
+    enable = (lib.mkEnableOption "Catppuccin globally") // {
+      default = config.catppuccin.enable;
+    };
```
getchoo, line 11:
```suggestion
    enable = lib.mkEnableOption "Catppuccin globally";
```
pretty sure this is also infinite recursion :p

getchoo (changes requested):

modules/nixos/globals.nix
pluiedev: It is definitely infinite recursion :P

modules/home-manager/globals.nix
drupol: Fixed

docs/home-manager-options.md
drupol: What do you mean ?

isabelroses: > What do you mean ?

the docs should not be updated in this PR, they are to be updated by ci upon the PR being merged. So discard the changes/.

drupol: OK !

modules/home-manager/globals.nix
```diff
@@ -11,5 +11,9 @@
       default = "teal";
       description = "Global Catppuccin accent";
     };
+
+    enable = (lib.mkEnableOption "Catppuccin globally") // {
+      default = config.catppuccin.enable;
+    };
```
isabelroses, line 17: is this not infinite recursion
```suggestion
    enable = lib.mkEnableOption "Catppuccin globally";
```

drupol: That works fine here.

drupol: Removing it doesn't have any impact, I'll remove it.

drupol:
Anything else to do?

modules/home-manager/globals.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, ... }: {
+{ lib, config, ... }: {
```
pluiedev, line 1: Unnecessary import

pluiedev (approved):
Just one small thing

modules/home-manager/globals.nix
drupol: True. Fixed.

pluiedev (approved):

getchoo (approved):
thanks!

becknik:
Thank you so much! This really feels like a enhancement to this project
I think documenting it in the `README` would be nice tho

Issue #103, opened 2024-04-09
Use themes from the cattpuccin package?


toast003:
Nixpkgs has a [catppuccin package](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/data/themes/catppuccin/default.nix) that has some of the themes this module can set, so it might be a good idea to use it if possible

toast003:
Got it

PR #129, opened 2024-04-18
feat(modules): add `catppuccin.sources` option

### LINKED ISSUE #96: expose palette to lib

getchoo:
fixes #96. sources for each port are now exposed and configurable under `catppuccin.sources` for both nixos and home-manager modules.

for contributors: this replaces nvfetcher with [npins](https://github.com/andir/npins) as the way we manage our port sources. this was to primarily allow for user defined sources to work as expected in our modules - as unlike nvfetcher, npins is a basic name value pair of the port name and path in the store

TODO:
- [ ] Document examples
- [x] Properly test
- [ ] Allow for interop with nixpkgs' sources as suggested in https://github.com/catppuccin/nix/issues/103 ?
  - this will be a simple PR, probably just adding their sources to `passthru` so they can be individually plugged in here


getchoo:
let's hope this doesn't break something 🙃

PR #133, opened 2024-04-19
feat(home-manager): add support for waybar

### LINKED ISSUE #126: Port Request: waybar
https://github.com/catppuccin/waybar

Lichthagel:
This currently only works if style is defined inline and not a separate css file.

Closes #126

### DIFF
--- a/modules/home-manager/waybar.nix
+++ b/modules/home-manager/waybar.nix
@@ -1,0 +1,18 @@
+{ config
+, lib
+, sources
+, ...
+}:
+let
+  cfg = config.programs.waybar.catppuccin;
+  enable = cfg.enable && config.programs.waybar.enable;
+in
+{
+  options.programs.waybar.catppuccin = lib.ctp.mkCatppuccinOpt "waybar";
+
+  config.programs.waybar = lib.mkIf enable {
+    style = lib.mkBefore ''
+      @import "${sources.waybar}/themes/${cfg.flavour}.css";
+    '';
+  };
+}
--- a/modules/home-manager/waybar.nix
+++ b/modules/home-manager/waybar.nix
@@ -17,7 +17,12 @@ in
         "createLink"
       ];
       default = "prependImport";
-      description = "Whether to prepend an `@import` (requires that `programs.waybar.style` is set to a string) or to create a symlink `~/.config/waybar/catppuccin.css` (requires you to import manually).";
+      description = ''
+        Defines how to include the catppuccin theme css file:
+
+        - `prependImport`: Prepends the import statement, if `programs.waybar.style` is a string (with default override priority).
+        - `createLink`: Creates a symbolic link `~/.config/waybar/catppuccin.css`, which needs to be included in the waybar stylesheet.
+      '';
     };
   };
 
--- a/modules/home-manager/waybar.nix
+++ b/modules/home-manager/waybar.nix
@@ -7,6 +7,7 @@
 let
   cfg = config.programs.waybar.catppuccin;
   enable = cfg.enable && config.programs.waybar.enable;
+  styleFile = "${sources.waybar}/themes/${cfg.flavour}.css";
 in
 {
   options.programs.waybar.catppuccin = (lib.ctp.mkCatppuccinOpt "waybar") // {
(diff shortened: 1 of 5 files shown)


Lichthagel:
Just added two options to control the module's behavior.

modules/home-manager/waybar.nix
```diff
@@ -0,0 +1,40 @@
...
+  cfg = config.programs.waybar.catppuccin;
+  enable = cfg.enable && config.programs.waybar.enable;
+in
+{
+  options.programs.waybar.catppuccin = (lib.ctp.mkCatppuccinOpt "waybar") // {
+    styleFile = lib.mkOption {
+      type = lib.types.path;
+      default = "${sources.waybar}/themes/${cfg.flavour}.css";
+      description = "Path to CSS file containing color variable definitions";
+    };
```
getchoo, line 16: im a bit hesitant on allowing this to be an option. no other module has this, and i feel like removing the file from the port would kinda defeat the purpose of using our module in the first place - as without it, you don't have a catppuccin theme. using alternative versions of this file could also be covered under https://github.com/catppuccin/nix/pull/129

modules/home-manager/waybar.nix
```diff
@@ -0,0 +1,40 @@
...
+      type = lib.types.bool;
+      default = false;
+      description = "Whether to prepend `@import \"$${styleFile}\";` when `programs.waybar.style` is set to a string.";
+    };
+
+    createLink = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = "Whether to create a symlink `~/.config/waybar/catppuccin.css` pointing to `styleFile`.";
+    };
```
getchoo, line 28: i think one of these should be `true` by default, as all other modules will apply something when enabled individually or with the new global option; this is essentially no-op with `catppuccin.enable = true`. i also don't see a reason why both of these should ever be enabled together, so an assertion would be good (unless i'm not thinking of some specific case. i don't use waybar personally lol)

getchoo (changes requested):

modules/home-manager/waybar.nix
Lichthagel: Yes, #129 definitely is the better solution once it hits. I removed the option for now.

modules/home-manager/waybar.nix
Lichthagel: Replaced these by an enum.

Lichthagel: Also added a warning if `prependImport` has no effect. (Can replace with an assertion if you prefer that.)

modules/home-manager/waybar.nix
```diff
@@ -0,0 +1,47 @@
...
+      description = "Whether to prepend an `@import` (requires that `programs.waybar.style` is set to a string) or to create a symlink `~/.config/waybar/catppuccin.css` (requires you to import manually).";
+    };
+  };
+
+  config = lib.mkIf enable (
+    let
+      styleFile = "${sources.waybar}/themes/${cfg.flavour}.css";
+    in
+    {
+      warnings =
```
getchoo, line 28: i don't think this warning is needed

`lib.modules.defaultOverridePriority` is currently `100`, which is equal to `lib.mkBefore`'s priority. this makes it so that under the precondition of the `"prependImport"` mode being used, this will never trigger. `lib.mkBefore` can also be used freely in the final configuration, with those entries still being placed after/having a higher priority than ours.

modules/home-manager/waybar.nix
getchoo, line 1: leaving this as a regular comment as the diff would be god awful

`mkMerge` can be used here to allow for a better separation between each mode (and make reading this a bit easier imo), and the description can use a touch up
```nix
{ config
, lib
, sources
, ...
}:
let
  cfg = config.programs.waybar.catppuccin;
  enable = cfg.enable && config.programs.waybar.enable;

  styleFile = "${sources.waybar}/themes/${cfg.flavour}.css";
in
{
  options.programs.waybar.catppuccin = lib.ctp.mkCatppuccinOpt "waybar" // {
    mode = lib.mkOption {
      type = lib.types.enum [
        "prependImport"
        "createLink"
      ];
      default = "prependImport";
      description = ''
        Whether to prepend an `@import` statement with the theme or to create a symlink targeting it at `~/.config/waybar/catppuccin.css` (requires you to import manually).
      '';
    };
  };

  config = lib.mkIf enable (
    lib.mkMerge [
      (lib.mkIf (cfg.mode == "prependImport") {
        programs.waybar.style = lib.mkBefore ''
          @import "${styleFile}";
        '';
      })

      (lib.mkIf (cfg.mode == "createLink") {
        xdg.configFile."waybar/catppuccin.css" = {
          source = styleFile;
        };
      })
    ]
  );
}
```

getchoo (changes requested):

modules/home-manager/waybar.nix
Lichthagel: Unless someone uses `lib.mkForce` (or similar). Then this will also override the definition in the module.

Vice versa `lib.mkForce (lib.mkBefore "...")` would override any definition using the default override priority (or higher value than `mkForce`'s 50).

And something like `lib.mkOverride options.programs.waybar.style.highestPrio (...)` is not possible due to infinite recursion (and should not be used anyway, as it can't be overridden).

modules/home-manager/waybar.nix
Lichthagel: Done. Let me know if you like the new description.

modules/home-manager/waybar.nix
Lichthagel: In some sense this is just me thinking that `lib.mkBefore` does prepend something to an option (regardless of priority), but it actually is just a definition that should be merged first.^^

So I can perfectly understand if we want to remove it, since users should be aware that `lib.mkForce` overrides things.

getchoo: yeah to me this is intended behavior. if the only situation this really helps in is with the use of `mkForce`, i feel like that could go for all modules here (which we don't have warnings for)

Lichthagel: removed the warning

getchoo (approved):
thanks!

PR #137, opened 2024-04-21
feat(tmux): add support for extraConfig and custom plugins

vdbe:
This PR adds 2 new options to the tmux module
1. extraConfig
2. plugin

## extraConfig
`programs.tmux.extraConfig` is placed after `programs.tmux.plugins` because of this there was no easy way to configure the plugin.
image

The plugin is ran on line 59 any options changed after will have no effect.

## plugin
Many people fork and customize the tmux plugin.


modules/home-manager/tmux.nix
```diff
@@ -18,12 +19,33 @@ let
...
 {
   options.programs.tmux.catppuccin =
-    lib.ctp.mkCatppuccinOpt "tmux";
+    ctp.mkCatppuccinOpt "tmux"
+    // {
+      plugin = mkOption {
+        type = types.package;
+        default = plugin;
+        description = "Catppuccin tmux plugin";
+      };
```
getchoo, line 28:
```suggestion

```

See my comment [here](https://github.com/catppuccin/nix/pull/133#discussion_r1576653324)

tl;dr: Replacing each theme source should be done in #129, not on a per module basis

modules/home-manager/tmux.nix
```diff
@@ -18,12 +19,33 @@ let
...
   config.programs.tmux.plugins = lib.mkIf enable [
     {
-      inherit plugin;
-      extraConfig = "set -g @catppuccin_flavour '${cfg.flavour}'";
+      inherit (cfg) plugin;
+      extraConfig = concatStrings [
+        ''
+          set -g @catppuccin_flavour '${cfg.flavour}'
+        ''
+        cfg.extraConfig
```
getchoo, line 47:
```suggestion
      extraConfig = lib.mkBefore ''
        set -g @catppuccin_flavour '${cfg.flavour}'
      '';
```
i don't think we really need to be re-exporting our parent option here either. `lib.mkBefore` can cover basically the same use case

getchoo (changes requested):

modules/home-manager/tmux.nix
vdbe: The issue with `mkBefore` is that `programs.tmux.plugins` a list is. An I don't know any ~~good~~ way of modifying that `extraConfig`. Is there a way to modify it?

getchoo: you're right, i must have had a brain fart there. this should be fine :)

getchoo (approved):
thanks!

PR #138, opened 2024-04-22
feat(home-manager): add transparent option for k9s

henrisota:
Adds option `transparent` to use the transparent version of the variants. By default, the non-transparent variants override the background colors.

### DIFF
--- a/modules/home-manager/k9s.nix
+++ b/modules/home-manager/k9s.nix
@@ -7,22 +7,27 @@ let
   cfg = config.programs.k9s.catppuccin;
   enable = cfg.enable && config.programs.k9s.enable;
 
-  themeFile = "catppuccin-${cfg.flavour}.yaml";
+  themeName = "catppuccin-${cfg.flavour}" + lib.optionalString cfg.transparent "-transparent";
+  themeFile = "${themeName}.yaml";
   themePath = "/skins/${themeFile}";
   theme = sources.k9s + "/dist/${themeFile}";
 in
 {
   options.programs.k9s.catppuccin =
-    lib.ctp.mkCatppuccinOpt "k9s";
-
-  config = lib.mkIf enable
-    {
-      assertions = [
-        (lib.ctp.assertXdgEnabled "k9s")
-      ];
-
-      xdg.configFile."k9s${themePath}".source = theme;
-
-      programs.k9s.settings.k9s.ui.skin = "catppuccin-${cfg.flavour}";
+    lib.ctp.mkCatppuccinOpt "k9s"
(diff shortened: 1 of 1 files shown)


modules/home-manager/k9s.nix
```diff
@@ -7,22 +7,43 @@ let
...
 
-  themeFile = "catppuccin-${cfg.flavour}.yaml";
+  themeFile =
+    "catppuccin-${cfg.flavour}"
+    + (
+      if cfg.transparent
+      then "-transparent"
+      else ""
+    )
+    + ".yaml";
```
getchoo, line 17: we should be able to use `lib.optionalString` here instead. this name can also be used later
```suggestion
  themeName = "catppuccin-${cfg.flavour}" + lib.optionalString cfg.transparent "-transparent";
  themeFile = "${themeName}.yaml";
```

modules/home-manager/k9s.nix
```diff
@@ -7,22 +7,43 @@ let
...
 in
 {
   options.programs.k9s.catppuccin =
-    lib.ctp.mkCatppuccinOpt "k9s";
+    lib.ctp.mkCatppuccinOpt "k9s"
+    // {
+      transparent = lib.mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = "Catppuccin transparent version for flavour";
```
getchoo, line 28:
```suggestion
      transparent = lib.mkEnableOption "transparent version of flavour";
```

modules/home-manager/k9s.nix
```diff
@@ -7,22 +7,43 @@ let
...
 
-      programs.k9s.settings.k9s.ui.skin = "catppuccin-${cfg.flavour}";
-    };
+        programs.k9s.settings.k9s.ui.skin =
+          "catppuccin-${cfg.flavour}"
+          + (
+            if cfg.transparent
+            then "-transparent"
+            else ""
+          );
```
getchoo, line 47:
```suggestion
        programs.k9s.settings.k9s.ui.skin = themeName;
```

getchoo (changes requested):

modules/home-manager/k9s.nix
henrisota: Change made.

modules/home-manager/k9s.nix
henrisota: Change made.

modules/home-manager/k9s.nix
henrisota: Change made.

PR #139, opened 2024-04-22
feat(home-manager): add support for zellij

eljamm:
Catppuccin is included in zellij and can be enabled as the [official source](https://github.com/catppuccin/zellij?tab=readme-ov-file#usage) instructs.

Screenshot from 2024-04-22 16-31-34


getchoo (approved):
thanks :)

PR #140, opened 2024-04-22
ci: allow doc updates without PR

sgoudham:


.github/workflows/update-docs.yml
getchoo, line 1: the permissions field should also be removed, as the runner token doesn't need write perms anymore

getchoo (changes requested):

getchoo (approved):

PR #144, opened 2024-04-27
feat(home-manager/fcitx5): add apply option

Lichthagel:
The current implementation overrides the configuration for fcitx5 classical ui. This prevents users from changing different ui settings apart from the theme. This adds the ability to optionally install the theme without applying.

### DIFF
--- a/modules/home-manager/fcitx5.nix
+++ b/modules/home-manager/fcitx5.nix
@@ -9,7 +9,16 @@ let
   enable = cfg.enable && config.i18n.inputMethod.enabled == "fcitx5";
 in
 {
-  options.i18n.inputMethod.fcitx5.catppuccin = ctp.mkCatppuccinOpt "Fcitx5";
+  options.i18n.inputMethod.fcitx5.catppuccin = ctp.mkCatppuccinOpt "Fcitx5" // {
+    apply = lib.mkOption {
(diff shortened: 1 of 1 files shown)


modules/home-manager/fcitx5.nix
```diff
@@ -9,7 +9,15 @@ let
   enable = cfg.enable && config.i18n.inputMethod.enabled == "fcitx5";
 in
 {
-  options.i18n.inputMethod.fcitx5.catppuccin = ctp.mkCatppuccinOpt "Fcitx5";
+  options.i18n.inputMethod.fcitx5.catppuccin = ctp.mkCatppuccinOpt "Fcitx5" // {
+    apply = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Applies the theme by overwriting `$XDG_CONFIG_HOME/fcitx5/conf/classicui.conf`.
```
getchoo, line 17:
```suggestion
        Applies the theme by overwriting `$XDG_CONFIG_HOME/fcitx5/conf/classicui.conf`.
        If this is disabled, you must manually set the theme
```

getchoo (changes requested):

getchoo (approved):
thanks!

PR #145, opened 2024-04-28
chore(modules): update flavor and accent defaults

uncenter:
Title.


Lichthagel:
Maybe we should also alias flavor to flavour.

uncenter:
> Maybe we should also alias flavor to flavour.

I've mentioned that before in the Discord and I think that is planned but not for this PR.

getchoo (approved):
thanks!

PR #147, opened 2024-04-29
fix: make dark/light lowecase for style names in gtk

selfuryon:
I tried to use Latte flavour, but all window titles are black. The reason of that I think is here: https://github.com/catppuccin/gtk/issues/100, so we need to use `dark` and `light` instead of `Dark` and `Light` in style.

current style:
image

fixed style:
image


getchoo (approved):
thanks :)

ecarlson94:
I believe this actually broke dark mode on my system.

jnsgruk:
yeh, see #159 which should fix it - though clearly there is some conflict here.

PR #149, opened 2024-04-29
feat(home-manager): add support for swaync

Lichthagel:
Ready-to-use css files are only provided as release artifacts.
We could also build these ourselves, but would need to manually update the `npmDepsHash`.

The default font used in the theme is "Ubuntu Nerd Font".
Adding it to `home.packages` is not possible as I query that for `enable` and thus would result in infinite recursion.

For the option name I referred to https://github.com/nix-community/home-manager/pull/4249 and https://github.com/nix-community/home-manager/pull/5345.

### DIFF
--- a/modules/home-manager/swaync.nix
+++ b/modules/home-manager/swaync.nix
@@ -1,0 +1,53 @@
+{ catppuccinLib }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.catppuccin.swaync;
+
+  enable = cfg.enable && config.services.swaync.enable;
+
+  theme = pkgs.substitute {
+    src = sources.swaync + "/${cfg.flavor}.css";
+
+    substitutions = [
+      "--replace-warn"
+      "Ubuntu Nerd Font"
+      cfg.font
+    ];
+  };
+in
+
+{
+  options.catppuccin.swaync =
+    catppuccinLib.mkCatppuccinOption {
+      name = "swaync";
+    }
+    // {
+      font = lib.mkOption {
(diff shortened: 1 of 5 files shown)


Lichthagel:
Rebased and adjusted for the now merged hm module.

Also installs the default font if it is selected.

Regarding the failing "Validate sources" CI: In `checkSources.nix` we might want to replace `isFromOrg` by 
```nix
isFromOrg = v: (v.src ? owner && v.src.owner == "catppuccin") || (builtins.match "https://github\.com/catppuccin/.*" v.src.url != null);
```
if we want to allow `fetchurl` for e.g. release artifacts.

modules/home-manager/swaync.nix
```diff
@@ -0,0 +1,38 @@
...
+
+    substitutions = [
+      "--replace-warn"
+      "Ubuntu Nerd Font"
+      cfg.font
+    ];
+  };
+in
+{
+  options.services.swaync.catppuccin = lib.ctp.mkCatppuccinOpt "sway" // {
```
isabelroses, line 22:
```suggestion
  options.services.swaync.catppuccin = lib.ctp.mkCatppuccinOpt "swaync" // {
```

getchoo:
going to wait on https://github.com/andir/npins/pull/46

Brisingr05:
The npins PR is merged.

Lichthagel:
> The npins PR is merged.

Despite the name, the PR only adds support for submodules.

isabelroses:
This should now be possible!

getchoo:
Just rebased and cleaned some things up. Should be ready :)

isabelroses (dismissed):

getchoo (approved):
Thanks!

PR #150, opened 2024-04-29
fix(home-manager/delta): use local flavour option

Lichthagel:


getchoo (approved):
i can't believe this has been here for so long lmfao
nice catch!

PR #151, opened 2024-04-29
chore(sources/gtk): remove dangling source

Lichthagel:
The gtk home-manager module uses catppuccin-gtk packaged in nixpkgs. This source is not used.


getchoo (approved):
thanks!

PR #161, opened 2024-05-04
feat(home-manager/gtk): add gnome option

c-leri:
Add an option to set the gtk theme as gnome shell's theme using the user-themes gnome extension

I'm not sure if this is the right way to declare this type of option or if I should have opted for something like `gnome.enable` instead so tell me what you think

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -25,6 +25,11 @@ in
         default = [ "normal" ];
         description = "Catppuccin tweaks for gtk";
       };
+      gnome = mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = "Wether to use the gtk theme as gnome shell's theme";
+      };
 
       cursor = ctp.mkCatppuccinOpt "gtk cursors"
       // {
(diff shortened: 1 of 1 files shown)


modules/home-manager/gtk.nix
```diff
@@ -25,6 +25,11 @@ in
         default = [ "normal" ];
         description = "Catppuccin tweaks for gtk";
       };
+      gnome = mkOption {
+        type = lib.types.bool;
+        default = false;
+        description = "Wether to use the gtk theme as gnome shell's theme";
+      };
```
getchoo, line 32:
```suggestion
      gnomeShellTheme = mkEnableOption "the Catppuccin GTK theme for GNOME Shell"
```
this could also just be `shellTheme` maybe? i'll leave that up to you

modules/home-manager/gtk.nix
```diff
@@ -79,5 +84,25 @@ in
         "gtk-4.0/gtk.css".source = "${gtk4Dir}/gtk.css";
         "gtk-4.0/gtk-dark.css".source = "${gtk4Dir}/gtk-dark.css";
       };
+
+    home.packages = lib.mkIf cfg.gnome [pkgs.gnomeExtensions.user-themes];
+
+    dconf.settings = lib.mkIf cfg.gnome {
```
getchoo, line 90: we should assert that dconf is enabled, similar to xdg above

getchoo (changes requested):
otherwise lgtm :)

modules/home-manager/gtk.nix
c-leri: I figured that since this is a theme for gnome, dconf would be enabled anyway but yes I can add that for good measure

c-leri: Actually dconf is enabled in the nixos config not in home-manager so I have no idea of how to assert that it is enabled in a home-manager module

getchoo: ah you're right, i forgot home-manager can't enable it. sorry about that :p

getchoo (approved):
thanks!

PR #168, opened 2024-05-13
feat(nixos): sddm

isabelroses:
very cool theme 0 bias

### DIFF
--- a/modules/nixos/sddm.nix
+++ b/modules/nixos/sddm.nix
@@ -1,0 +1,51 @@
+{ lib
+, pkgs
+, config
+, ...
+}:
+let
+  cfg = config.services.displayManager.sddm.catppuccin;
+  enable = cfg.enable && config.services.displayManager.sddm.enable;
+
+  inherit (lib) mkIf ctp types mkOption;
+in
+{
+  options.services.displayManager.sddm.catppuccin =
+    ctp.mkCatppuccinOpt "sddm"
+    // {
+      font = mkOption {
+        type = types.str;
+        default = "Noto Sans";
+        description = "Font to use for the login screen";
+      };
+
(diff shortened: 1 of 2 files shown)


isabelroses:
flake.lock needs to be bumped

getchoo (approved):
whoever's maintaining this theme is doing awesome. wonder who it is??

i think this is also going to keep me on plasma 6 for another week before going back to gnome on 46. thanks!

ecarlson94:
This doesn't work for me because of [this PR](https://github.com/catppuccin/sddm/pull/26). In fact, this only works for NixOS users that use Qt (which I don't). I had to remove this line to get the theme to work. This tells me that we should remove this line from the metadata file and manage Qt support within the install phase of the [catppuccin-sddm package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ca/catppuccin-sddm/package.nix) instead (it should be opt in with an option like `qt.enable`).

isabelroses:
> This doesn't work for me because of [this PR](https://github.com/catppuccin/sddm/pull/26). In fact, this only works for NixOS users that use Qt (which I don't). I had to remove this line to get the theme to work. This tells me that we should remove this line from the metadata file and manage Qt support within the install phase of the [catppuccin-sddm package](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ca/catppuccin-sddm/package.nix?rgh-link-date=2024-05-15T05%3A51%3A33Z) instead (it should be opt in with an option like `qt.enable`).

This seems like an issue you should take up with the upstream packager (nixpkgs). Since sddm is under the KDE banner so it makes sense for the catppuccin-sddm theme (and its maintainer, me) to ship it with the qt-platform set to 6 by default.

PR #179, opened 2024-05-20
feat(nixos): add support for gitea/forgejo

Lichthagel:
Requires fetching release artifacts or building the theme from source (similar to #149)

### DIFF
--- a/modules/nixos/gitea.nix
+++ b/modules/nixos/gitea.nix
@@ -1,0 +1,88 @@
+{ catppuccinLib }:
+{
+  lib,
+  config,
+  ...
+}:
+let
+  inherit (config.catppuccin) sources;
+
+  valuesFromEnum =
+    enum: if lib.isList enum.functor.payload then enum.functor.payload else enum.functor.payload.values;
+
+  supportedForges = [
+    "gitea"
+    "forgejo"
+  ];
+
+  builtinThemes = {
+    gitea = [
+      "auto"
+      "gitea"
+      "arc-greeen"
+    ];
+
+    forgejo = [
+      "forgejo-auto"
+      "forgejo-light"
+      "forgejo-dark"
+      "gitea-auto"
+      "gitea-light"
+      "gitea-dark"
+      "forgejo-auto-deuteranopia-protanopia"
+      "forgejo-light-deuteranopia-protanopia"
+      "forgejo-dark-deuteranopia-protanopia"
+      "forgejo-auto-tritanopia"
+      "forgejo-light-tritanopia"
+      "forgejo-dark-tritanopia"
+    ];
+  };
+in
+{
+  options.catppuccin = lib.genAttrs supportedForges (
+    name:
+    catppuccinLib.mkCatppuccinOption {
+      inherit name;
+      accentSupport = true;
+    }
+  );
+
+  config = lib.mkMerge (
+    map (
+      forge:
+      let
+        cfg = config.catppuccin.${forge};
+
+        inherit (config.services.${forge}) customDir;
+        themeDir =
+          if lib.versionAtLeast config.services.${forge}.package.version "1.21.0" then
+            "${customDir}/public/assets/css"
+          else
+            "${customDir}/public/css";
+      in
+      lib.mkIf cfg.enable {
+        systemd.tmpfiles.settings."10-catppuccin-${forge}-theme" = {
+          ${themeDir}."L+" = {
+            argument = toString sources.gitea;
+          };
+
+          ${dirOf themeDir}.d = {
+            inherit (config.services.${forge}) user group;
+          };
+        };
+
+        services.${forge}.settings.ui = {
+          DEFAULT_THEME = "catppuccin-${cfg.flavor}-${cfg.accent}";
+          THEMES = lib.concatStringsSep "," (
+            builtinThemes.${forge}
+            ++ (lib.mapCartesianProduct ({ flavor, accent }: "catppuccin-${flavor}-${accent}") (
+              lib.mapAttrs (lib.const valuesFromEnum) {
+                inherit (catppuccinLib.types) accent flavor;
+              }
+            ))
+          );
+        };
+      }
+    ) supportedForges
+  );
+}
--- a/pkgs/gitea/package.nix
+++ b/pkgs/gitea/package.nix
@@ -1,0 +1,6 @@
+{ fetchzip }:
+fetchzip {
+  url = "https://github.com/catppuccin/gitea/releases/download/v0.4.1/catppuccin-gitea.tar.gz";
+  hash = "sha256-14XqO1ZhhPS7VDBSzqW55kh6n5cFZGZmvRCtMEh8JPI=";
+  stripRoot = false;
+}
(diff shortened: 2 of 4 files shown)


isabelroses:
This should now be possible!

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,81 @@
...
+      "forgejo-auto-deuteranopia-protanopia"
+      "forgejo-light-deuteranopia-protanopia"
+      "forgejo-dark-deuteranopia-protanopia"
+      "forgejo-auto-tritanopia"
+      "forgejo-light-tritanopia"
+      "forgejo-dark-tritanopia"
+    ];
+  };
+
+  generateConfig =
```
getchoo, line 38: Nit, but I'd prefer this to be inlined below rather than it's own lambda

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,81 @@
...
+            if lib.versionAtLeast config.services.${forge}.package.version "1.21.0" then
+              "${customDir}/public/assets"
+            else
+              "${customDir}/public";
+        in
+        lib.mkAfter ''
+          rm -rf ${baseDir}/css
+          mkdir -p ${baseDir}
+          ln -sf ${sources.gitea} ${baseDir}/css
+        '';
```
getchoo, line 58: Why not uses systemd-tmpfiles?

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,81 @@
...
+
+      services.${forge}.settings.ui = {
+        DEFAULT_THEME = "catppuccin-${cfg.flavor}-${cfg.accent}";
+        THEMES = builtins.concatStringsSep "," (
+          builtinThemes.${forge}
```
getchoo, line 63: Do we really have to enumerate these? It doesn't seem like the port's example `app.ini` specifies them

pkgs/gitea/package.nix
```diff
@@ -0,0 +1,23 @@
+{ buildCatppuccinPort, fetchzip }:
+
+buildCatppuccinPort (finalAttrs: {
+  pname = "gitea";
+  version = "0.4.1";
+
+  src = fetchzip {
+    url = "https://github.com/catppuccin/gitea/releases/download/v${finalAttrs.version}/catppuccin-gitea.tar.gz";
+    sha256 = "sha256-14XqO1ZhhPS7VDBSzqW55kh6n5cFZGZmvRCtMEh8JPI=";
```
getchoo, line 9:
```suggestion
    hash = "sha256-14XqO1ZhhPS7VDBSzqW55kh6n5cFZGZmvRCtMEh8JPI=";
```

pkgs/gitea/package.nix
getchoo, line 1: Pretty sure this file could just be a `fetchzip` call, unless we really want that `meta` attribute from `buildCatppuccinPort`. Up to you

pkgs/gitea/package.nix
```diff
@@ -0,0 +1,23 @@
...
+    stripRoot = false;
+  };
+
+  dontCatppuccinInstall = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out
+    mv * $out
```
getchoo, line 19:
```suggestion
    ln -s $src $out
```
Pretty sure this would work. It's a bit neater too

modules/lib/default.nix
getchoo, line 1: Can't say I'm a fan of introducing a general `consts` just for this

Maybe access them through `types.flavors.functor.payload.values` instead?
Now that I type it out it does sound a bit silly, though...

modules/nixos/gitea.nix
isabelroses: Catppuccin doesn't list them but I know from my own instance that if we don't the user does not get the builtin themes available.

modules/lib/default.nix
isabelroses: I wasn't a fan either hence I wanted to prompt you for a review. Strictly speaking we could drop it again and use IFD. But I'd rather avoid introducing more IFD.

modules/nixos/gitea.nix
getchoo: Ah that sucks :(

modules/lib/default.nix
getchoo: I think `types.flavors.functor.payload.values` would be fine for now then, even though it's *super* ugly

If more ports end up using it later, I'd be fine with a `consts`

getchoo:
Sorry for the wait again, I've been setting up my own Forgejo instance for a little bit and figured it would be the perfect place to test this out. I'll hopefully get to it soon :p

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,78 @@
...
+        };
+
+        services.${forge}.settings.ui = {
+          DEFAULT_THEME = "catppuccin-${cfg.flavor}-${cfg.accent}";
+          THEMES = builtins.concatStringsSep "," (
+            builtinThemes.${forge}
+            ++ (lib.mapCartesianProduct ({ flavors, accents }: "catppuccin-${flavors}-${accents}") {
+              flavors = catppuccinLib.types.flavor.functor.payload.values;
+              accents = catppuccinLib.types.accent.functor.payload.values;
+            })
```
getchoo, line 72:
```suggestion
            ++ (lib.mapCartesianProduct ({ flavor, accent }: "catppuccin-${flavor}-${accent}") (
              lib.mapAttrs (lib.const valuesFromEnum) {
                inherit (catppuccinLib.types) accent flavor;
              }
            ))
```

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,78 @@
+{ catppuccinLib }:
+{
+  lib,
+  config,
+  ...
+}:
+let
+  inherit (config.catppuccin) sources;
```
getchoo, line 8: This goes unused...

modules/nixos/gitea.nix
```diff
@@ -0,0 +1,78 @@
...
+        themeDir =
+          if lib.versionAtLeast config.services.${forge}.package.version "1.21.0" then
+            "${customDir}/public/assets/css"
+          else
+            "${customDir}/public/css";
+      in
+      lib.mkIf cfg.enable {
+        systemd.tmpfiles.settings."10-catppuccin-${forge}-theme".${themeDir}."L+" = {
+          argument = themeDir;
+        };
```
getchoo, line 63: I'm *pretty sure* this is just making this directory a symlink to itself?

isabelroses: yeah i just noticed this should be `sources.gitea`

isabelroses:
> Sorry for the wait again, I've been setting up my own Forgejo instance for a little bit and figured it would be the perfect place to test this out. I'll hopefully get to it soon :p

On this note. Just deployed my git instance with the theme and it works perfectly.

getchoo (changes requested):
With the following applied it works on my Forgejo instance as well :)

getchoo (approved):
Thanks for your work @Lichthagel and @isabelroses. Sorry for the long wait, too

PR #194, opened 2024-05-24
fix(nixos): sddm package not being installed 

### LINKED ISSUE #192: SDDM Themeing broke
All i can say for sure is that before the renaming of "flavour" to "flavor" the themeing worked, and after i saw the error being caused by the rename and fixed them, then upgraded it no longer was themed. It looks like the default SDDM Theme, so not the Plasma6 theme or anything, just the default default one.

I will attach a screenshot below, just need to figure out how to do that part.

Weathercold:
Never use attrset merge operator with `mk*` attrsets since the right hand attrset overrides the left

Fixes #192

### DIFF
--- a/modules/nixos/sddm.nix
+++ b/modules/nixos/sddm.nix
@@ -13,12 +13,16 @@ let
     versionAtLeast
     ;
   cfg = config.services.displayManager.sddm.catppuccin;
-  enable = cfg.enable && config.services.displayManager.sddm.enable;
+  enable =
+    versionAtLeast ctp.getModuleRelease minVersion
+    && cfg.enable
+    && config.services.displayManager.sddm.enable;
 
   # versions >= 24.05 renamed `services.xserver.displayManager` to `services.displayManager`
   # TODO: remove when 24.05 is stable
   minVersion = "24.05";
 in
+
 {
   options.services.displayManager = ctp.mkVersionedOpts minVersion {
     sddm.catppuccin = ctp.mkCatppuccinOpt "sddm" // {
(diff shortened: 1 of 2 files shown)


Weathercold:
It seems that aef56729 tries to install the theme package for all versions, but set the theme only if >= 24.05. However, since the options are only made accessible if >= 24.05, there's no way to enable sddm theming at all on older versions, so I changed the module to not enable `config` on older versions

modules/nixos/sddm.nix
```diff
@@ -13,12 +13,16 @@ let
     versionAtLeast
     ;
   cfg = config.services.displayManager.sddm.catppuccin;
-  enable = cfg.enable && config.services.displayManager.sddm.enable;
+  enable =
+    versionAtLeast ctp.getModuleRelease minVersion
```
vdbe, line 17: Why is this extra version check needed? You already can't enable `cfg` on versions under `minVersion`.

Maybe it's better to place the `mkVersionedOpts` on `options.services.displayManager.sddm.catppuccin` so you get this error 
image
instead of this 
image
when trying to enable `services.displayManager.sddm` with a version under `minVersion`

getchoo (approved):
lgtm. sorry for the wait and thank you!

modules/nixos/sddm.nix
Weathercold: I didn't test it, but theoretically, without the version check, `enable` depends on `config.services.displayManager.sddm.catppuccin` even when it's not defined and not enabled

PR #195, opened 2024-05-24
feat(home-manager)!: add support for global cursors

### LINKED ISSUE #158: split cursors into separate module
as it is now, i have to apply them twice to get them working with all apps like [this](https://git.gay/pinkcreeper100/nix-config/src/branch/main/modules/home/catppuccin/default.nix#L23)

Weathercold:
Closes #158

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -6,27 +6,24 @@
 }:
 let
   inherit (lib)
+    concatStringsSep
     ctp
-    mkOption
+    mkIf
     mkEnableOption
+    mkMerge
+    mkOption
+    mkRenamedOptionModule
     types
     ;
   cfg = config.gtk.catppuccin;
   enable = cfg.enable && config.gtk.enable;
-  # "dark" and "light" can be used alongside the regular accents
-  cursorAccentType = ctp.mergeEnums ctp.types.accentOption (
-    lib.types.enum [
-      "dark"
-      "light"
-    ]
-  );
 in
 {
   options.gtk.catppuccin = ctp.mkCatppuccinOpt "gtk" // {
     # NOTE: we are overriding the previous declaration of `enable` here
     # as this module is deprecated and we do not want it to apply with
     # the global `catppuccin.enable`
-    enable = lib.mkEnableOption "Catppuccin theme";
+    enable = mkEnableOption "Catppuccin theme";
 
     accent = ctp.mkAccentOpt "gtk";
     size = mkOption {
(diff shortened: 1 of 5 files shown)


modules/home-manager/home.nix
```diff
@@ -0,0 +1,30 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  inherit (lib) ctp mkIf;
+  cfg = config.home.catppuccin;
```
getchoo, line 9:
```suggestion
  cfg = config.home.pointerCursor.catppuccin;
```
we usually place name our options after the main option we're changing

modules/home-manager/home.nix
```diff
@@ -0,0 +1,30 @@
...
+in
+{
+  options.home.catppuccin.cursor =
```
getchoo, line 12:
```suggestion
  options.home.pointerCursor.catppuccin =
```

getchoo (changes requested):

modules/home-manager/home.nix
Weathercold: I did try this, but `home.pointerCursor` is a module and I can't define extra options outside of the module

Weathercold: - `home.catppuccin.cursor`
	I still think this one is the best
 - `catppuccin.home.pointerCursor`
	 I don't like this one because `catppuccin` is only for global options
 - `home.pointerCursorCatppuccin`
	This one can work too
 - Send PR upstream to not define `home.pointerCursor` as a module
 	Too complex and might not even be possible
 - Move all `*.catppuccin` options to `catppuccin.*` and release new major version, breaking all existing configs
 	Probably too late to change to this approach, but it feels less hacky and more predictable since all options defined by our module are under the same attribute (`catppuccin`). The configuration can be concentrated to the same location in the users' config files. It's also how stylix defines its options.

isabelroses: perhaps define it under the `catppuccin.cursor` since `pointerCursor` is a "global" option in the same way... at least thats my thought process

modules/home-manager/home.nix
```diff
@@ -0,0 +1,30 @@
...
+      accent = ctp.mkCursorAccentOpt "global cursors";
+    };
+
+  config.home.pointerCursor =
+    let
+      flavorUpper = ctp.mkUpper cfg.cursor.flavor;
+      accentUpper = ctp.mkUpper cfg.cursor.accent;
+    in
+    mkIf cfg.cursor.enable {
+      name = "Catppuccin-${flavorUpper}-${accentUpper}-Cursors";
```
dixslyf, line 27: I believe the name was changed to all lowercase as of catppuccin cursors v0.2.1. Might be a good idea to set the name based on the version of the package.

dixslyf (changes requested):

modules/home-manager/home.nix
getchoo: i think `catppuccin.pointerCursor` would be a good option for the same reasons isabel mentioned, but also to keep parity with the `home.pointerCursor` option

getchoo: > Move all *.catppuccin options to catppuccin.* and release new major version, breaking all existing configs
Probably too late to change to this approach, but it feels less hacky and more predictable since all options defined by our module are under the same attribute (catppuccin). The configuration can be concentrated to the same location in the users' config files. It's also how stylix defines its options.

https://github.com/catppuccin/nix/issues/222

modules/home-manager/home.nix
Weathercold: we need to support 24.05 which is still on v0.2.0

isabelroses: um opsies... should of thought about that before https://github.com/catppuccin/nix/pull/212. sigh...

getchoo: https://github.com/NixOS/nixpkgs/pull/321759

getchoo:
i've adopted the new option naming in the `catppuccin` namespace as discussed above, and also made this a replacement for the original gtk module. this will resolve deprecation warnings for those only using that (as it's not actually deprecated or archived like the upstream gtk port) and prevent this PR from introducing another option that basically does the same thing as the old one

modules/home-manager/gtk.nix
```diff
@@ -48,31 +45,68 @@ in
...
 
     icon = ctp.mkCatppuccinOpt "gtk modified Papirus icon theme" // {
       accent = ctp.mkAccentOpt "gtk modified Papirus icon theme";
     };
   };
 
-  config = lib.mkIf enable {
-    warnings = [
+  imports = [
+    (mkRemovedOptionModule
```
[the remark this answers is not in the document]
getchoo: my original thought was that i wanted to not limit the interface of the new option to this one as to avoid breaking changes after 1.0.0...but after thinking about it more, i really doubt that'll happen since these will probably stay for a good while 😆

modules/home-manager/gtk.nix
```diff
@@ -83,49 +117,51 @@ in
...
-      let
-        gtk4Dir = "${config.gtk.theme.package}/share/themes/${config.gtk.theme.name}/gtk-4.0";
-      in
-      {
-        "gtk-4.0/assets".source = "${gtk4Dir}/assets";
-        "gtk-4.0/gtk.css".source = "${gtk4Dir}/gtk.css";
-        "gtk-4.0/gtk-dark.css".source = "${gtk4Dir}/gtk-dark.css";
-      };
+    (mkIf (enable && cfg.gnomeShellTheme) {
+      assertions = [
```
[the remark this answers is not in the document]
getchoo: thanks for catching this

..i need to stop writing modules at 3am

getchoo (approved):

getchoo:
thanks!

PR #201, opened 2024-05-29
ci(release-please): use action from googleapis

sgoudham:
I confirmed that it shouldn't break anything
from looking at the documentation, but who
knows with release-please /shrug


getchoo (approved):
lets hope it works lol

PR #203, opened 2024-05-30
feat(modules)!: bump minimum supported release to 24.05

getchoo:
draft until the actual release


Weathercold:
you should add a global assertion for the minimum nixos version

PR #206, opened 2024-06-04
chore(home-manager): deprecate gtk module

getchoo:
https://github.com/catppuccin/gtk/issues/262


getchoo:
#211 should be merged, updating the gtk port repo to its final revision

bbigras:
Is there an alternative to this module? my gnome (and apps) were so pretty with it.

Weathercold:
Are the cursor and icon configs deprecated too? These themes are not archived. When we eventually remove the options, we can just put a notice about the removal of the gtk theme in the description of `gtk.catppuccin`

DrymarchonShaun:
> Are the cursor and icon configs deprecated too? These themes are not archived. When we eventually remove the options, we can just put a notice about the removal of the gtk theme in the description of `gtk.catppuccin`

@getchoo I was wondering about this as well.

Airradda:
> Is there an alternative to this module? my gnome (and apps) were so pretty with it.

In terms of themes, there is [Fausto-Korpsvart/Catppuccin-GTK-Theme](https://github.com/Fausto-Korpsvart/Catppuccin-GTK-Theme) which has been recently updated and there is/was [talk of it merged into the Org](https://github.com/Fausto-Korpsvart/Catppuccin-GTK-Theme/issues/1#issuecomment-1188955595), but it requires the murrine-engine which I don't think has been packaged yet.

getchoo:
> Are the cursor and icon configs deprecated too? These themes are not archived.

no, these ports are not; however i would consider these *options* deprecated for the time being as i plan on splitting them into their own options (like in https://github.com/catppuccin/nix/pull/195). this should have been better communicated, that's my bad

> Is there an alternative to this module? my gnome (and apps) were so pretty with it.

i think this would be better left to a conversation in the gtk ports repository
but personally, i would recommend adw-gtk3 and gradience's catppuccin theme. it's worked well when i used it - though it's been a while :p

DrymarchonShaun:
> but personally, i would recommend adw-gtk3 and gradience's catppuccin theme. it's worked well when i used it - though it's been a while :p

Actually just did this a few days ago, it's a bit [janky](https://github.com/DrymarchonShaun/nix-config/blob/stable/home/shaun/common/optional/desktops/gtk.nix) and there's probably a better way to do it, but it works. Honestly, I think it looks better than the GTK theme.

Weathercold:
btw I wrote a module for https://github.com/Fausto-Korpsvart/Catppuccin-GTK-Theme at https://github.com/Weathercold/nixfiles/blob/master/home/modules/themes/catppuccin/gtk.nix. You can integrate it into your config with some modifications

PR #212, opened 2024-06-06
chore(home-manager): cursors are now lowercase

isabelroses:
lovely init 

ref https://github.com/NixOS/nixpkgs/commit/914a9ca2c8a0d2cc934bc8a63589346828cd2c75

PR #218, opened 2024-06-10
feat(home-manager): set hyprcursor

isabelroses:
Implements my suggestion from https://github.com/catppuccin/nix/pull/212#issuecomment-2153410194

Felt like it would be good as a separate PR.

### DIFF
--- a/modules/home-manager/hyprland.nix
+++ b/modules/home-manager/hyprland.nix
@@ -3,19 +3,34 @@ let
   inherit (config.catppuccin) sources;
   cfg = config.wayland.windowManager.hyprland.catppuccin;
   enable = cfg.enable && config.wayland.windowManager.hyprland.enable;
+  inherit (config.gtk.catppuccin) cursor;
 in
 {
   options.wayland.windowManager.hyprland.catppuccin = lib.ctp.mkCatppuccinOpt "hyprland" // {
     accent = lib.ctp.mkAccentOpt "hyprland";
   };
 
-  config.wayland.windowManager.hyprland.settings = lib.mkIf enable {
-    source = [
-      "${sources.hyprland}/themes/${cfg.flavor}.conf"
-      (builtins.toFile "hyprland-${cfg.accent}-accent.conf" ''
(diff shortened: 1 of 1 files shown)


modules/home-manager/gtk.nix
```diff
@@ -105,7 +105,14 @@ in
...
       };
 
-    home.packages = lib.mkIf cfg.gnomeShellTheme [ pkgs.gnomeExtensions.user-themes ];
+    home = {
+      packages = lib.mkIf cfg.gnomeShellTheme [ pkgs.gnomeExtensions.user-themes ];
+
+      sessionVariables = lib.mkIf cfg.cursor.enable {
+        HYPRCURSOR_SIZE = "${toString cfg.size}";
+        HYPRCURSOR_THEME = "catppuccin-${cfg.cursor.flavor}-${cfg.cursor.accent}-cursors";
+      };
```
getchoo, line 114: i would prefer to keep this in the hyprland module, as there's not much reason for these to be set by default when only the gtk cursor is enabled

getchoo (changes requested):

modules/home-manager/hyprland.nix
```diff
@@ -3,19 +3,33 @@ let
...
+          $accent=''$${cfg.accent}
+          $accentAlpha=''$${cfg.accent}Alpha
+        '')
+        (lib.mkIf cursor.enable (
+          builtins.toFile "hyprland-cursors.conf" ''
+            env = HYPRCURSOR_THEME,MyCursor
+            env = HYPRCURSOR_SIZE,24
+          ''
+        ))
+      ];
```
getchoo, line 32:
```suggestion
      ] ++ lib.optional cursor.enable (
        builtins.toFile "hyprland-cursors.conf" ''
          env = HYPRCURSOR_THEME,MyCursor
          env = HYPRCURSOR_SIZE,24
        ''
      );
```

getchoo (changes requested):

getchoo (approved):
thanks!

PR #219, opened 2024-06-10
fix(home-manager): remove incorrect config in gitui module

Zh40Le1ZOOB:


getchoo:
i am a bit confused here, as this is what the upstream port does as shown [here](https://github.com/catppuccin/yazi/blob/0846aed69b2a62d29c98e100af0cf55ca729723d/themes/mocha.toml#L32) (also see #111). could you explain why we might not want this?

isabelroses:
Also that this is not the "gitui" module. Perhaps you made a PR from the wrong branch or something?

PR #220, opened 2024-06-10
chore(nixos): use theme package in nixpkgs for grub

Zh40Le1ZOOB:


getchoo:
the use of the grub port's source directly is intentional, as we prefer to maintain and update ports at our own pace (which is hopefully faster than nixpkgs) and it helps avoid any issues with users having a different revision of nixpkgs than the one we've tested with. thanks though :)

PR #221, opened 2024-06-10
feat(nixos): add support for fcitx5

Zh40Le1ZOOB:


modules/nixos/fcitx5.nix
```diff
@@ -0,0 +1,19 @@
...
+let
+  inherit (lib) ctp mkIf;
+  cfg = config.i18n.inputMethod.fcitx5.catppuccin;
+  enable = cfg.enable && config.i18n.inputMethod.enabled == "fcitx5";
+in
+{
+  options.i18n.inputMethod.fcitx5.catppuccin = ctp.mkCatppuccinOpt "fcitx5";
+
+  config.i18n.inputMethod.fcitx5 = mkIf enable {
+    addons = [ pkgs.fcitx5-catppuccin ];
```
getchoo, line 16: we should be using our own `fcitx5` source like we do in the home-manager module

getchoo (changes requested):

Zh40Le1ZOOB:
Hi, any update?

Weathercold (changes requested):
You should add `i18n.inputMethod.enabled = "fcitx5";` in the nixos test

Zh40Le1ZOOB:
> You should add `i18n.inputMethod.enabled = "fcitx5";` in the nixos test

Thank you for review, done.

modules/nixos/fcitx5.nix
```diff
@@ -0,0 +1,29 @@
...
+  ...
+}:
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.i18n.inputMethod.fcitx5.catppuccin;
+  enable =
+    cfg.enable
+    && (
+      config.i18n.inputMethod.enabled == "fcitx5" # Deprecated but still available
+      || (config.i18n.inputMethod.enable && config.i18n.inputMethod.type == "fcitx5")
```
Weathercold, line 14: we need to support 24.05 which doesn’t have `config.i18n.inputMethod.enable`. Evaluation of `config.i18n.inputMethod.enabled` should be avoided since it will be removed in the future. Do

```nix
cfg.enable
&& (
  config.i18n.inputMethod ? enable
  && config.i18n.inputMethod.enable
  && config.i18n.inputMethod.type == "fcitx5"
  || (!config.i18n.inputMethod ? enable && config.i18n.inputMethod.enabled == "fcitx5")
)
```

Weathercold (changes requested):
Add your module to `modules/nixos/all-modules.nix`

modules/nixos/fcitx5.nix
```diff
@@ -0,0 +1,31 @@
...
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.i18n.inputMethod.fcitx5.catppuccin;
+  enable =
+    cfg.enable
+    && (
+      config.i18n.inputMethod ? enable
+      && config.i18n.inputMethod.enable
+      && config.i18n.inputMethod.type == "fcitx5"
+      || (!config.i18n.inputMethod ? enable && config.i18n.inputMethod.enabled == "fcitx5")
```
getchoo, line 16: I'm pretty sure this can be simplified to the following. 
```suggestion
      config.i18n.inputMethod.enable or true
      && config.i18n.inputMethod.type or config.i18n.inputMethod.enabled == "fcitx5"
```
The first condition is the check for the newer 24.11 option, but should be ok to fall back on it being true as the next option will check the main `enabled` option from 24.05

getchoo (changes requested):
Sorry for the wait!

modules/nixos/fcitx5.nix
Weathercold: Right, they should be identical. clever

Zh40Le1ZOOB: Thank you very much!

getchoo (changes requested):
These reviews have apparently been pending for about a month...guess I forgot to hit submit 🤦🏻

Weathercold (changes requested):
You should add accent support (see #343 above), otherwise LGTM

Zh40Le1ZOOB:
> You should add accent support (see #343 above), otherwise LGTM

Done, thanks for review.

modules/nixos/fcitx5.nix
```diff
@@ -0,0 +1,31 @@
...
+  enable =
+    cfg.enable
+    && (
+      config.i18n.inputMethod.enable or true
+      && config.i18n.inputMethod.type or config.i18n.inputMethod.enabled == "fcitx5"
+    );
+
+  theme = pkgs.runCommand "catppuccin-fcitx5" { } ''
+    mkdir -p $out/share/fcitx5/themes/
+    cp -r ${sources.fcitx5}/src/catppuccin-${cfg.flavor}/ $out/share/fcitx5/themes/
```
arminius-smh, line 19: the path of the themes got updated as well and is `catppuccin-${cfg.flavor}-${cfg.accent}` now

getchoo (approved):
Thanks and sorry again for the wait!

Zh40Le1ZOOB:
> Thanks and sorry again for the wait!

Thank you very much!

PR #75, opened 2024-01-28
feat(home-manager): add support for fuzzel

Anomalocaridid:
Adds support for fuzzel.

PR #224, opened 2024-06-11
feat(home-manager): add support for fuzzel

SchweGELBin:
Added support for fuzzel.
Used ```xdg.configFile```, maybe there is a way to convert it to ```programs.fuzzel.settings.colors```.


SchweGELBin:
Sorry, I wasn't aware of #75.

SchweGELBin:
Now the border color is the accent color (with the help of catppuccin/palette).

getchoo:
i would prefer to go with #75 as that (primarily) uses the upstream port over the full palette (which we consider to be a better practice)

SchweGELBin:
I'm good with it, made this pr, because I didn't know it existed. If I can help, let me know.

getchoo:
i've gone back and rebased the first PR, so closing this in favor of it

Issue #229, opened 2024-06-14
SDDM theme not working


jerbaroo:
I have the global NixOS Catppuccin flag enabled: https://github.com/jerbaroo/nixos/blob/ecdfb7a2982d13f0ba2353f3fc9fa2287bea4ac3/system/theme.nix#L4

And SDDM is also enabled: https://github.com/jerbaroo/nixos/blob/ecdfb7a2982d13f0ba2353f3fc9fa2287bea4ac3/system/graphics.nix#L11

However SDDM looks the same as it does in the other issue about SDDM being broken (#192).

My `catppuccin-nix` version is: [d34a94a17c6ec4a0c4e24b3e4336ea504d021f6d](https://github.com/jerbaroo/nixos/blob/ecdfb7a2982d13f0ba2353f3fc9fa2287bea4ac3/flake.lock#L27) 

```
> cat /etc/sddm.conf
[General]
DefaultSession=
DisplayServer=wayland
HaltCommand=/run/current-system/systemd/bin/systemctl poweroff
Numlock=none
RebootCommand=/run/current-system/systemd/bin/systemctl reboot

[Theme]
Current=catppuccin-mocha
FacesDir=/run/current-system/sw/share/sddm/faces
ThemeDir=/run/current-system/sw/share/sddm/themes

[Users]
HideShells=/run/current-system/sw/bin/nologin
HideUsers=nixbld1,nixbld10,nixbld11,nixbld12,nixbld13,nixbld14,nixbld15,nixbld16,nixbld17,nixbld18,nixbld19,nixbld2,nixbld20,nixbld21,nixbld22,nixbld23,nixbld24,nixbld25,nixbld26,nixbld27,nixbld28,nixbld29,nixbld3,nixbld30,nixbld31,nixbld32,nixbld4,nixbld5,nixbld6,nixbld7,nixbld8,nixbld9
MaximumUid=30000

[Wayland]
CompositorCommand=/nix/store/xzx28k96shd3l886s676iq01ngjkwi8n-weston-13.0.1/bin/weston --shell=kiosk -c /nix/store/za8lq4j5zlrln8l8lzvyylb4z4ngcbnq-weston.ini
EnableHiDPI=true
SessionDir=/nix/store/wg9x90w2fiy75mdczp46959ykpj638p1-desktops/share/wayland-sessions

```

isabelroses:
Can you test if https://github.com/catppuccin/nix/pull/230 fixes the issue?

[...]

PR #230, opened 2024-06-14
fix(nixos/sddm): use the kde sddm package

isabelroses:
This resolves the error where the theme doesn't work if plasma6 isn't enabled. by default the packge is qt5 sddm, but catppuccin/sddm uses qt6.

### DIFF
--- a/modules/nixos/sddm.nix
+++ b/modules/nixos/sddm.nix
@@ -40,10 +40,32 @@ in
       default = true;
       description = "Add an additional background layer to the login panel";
     };
+
+    assertQt6Sddm =
+      lib.mkEnableOption ''
+        checking if `services.displayManager.sddm.package` is the Qt 6 version.
+
+        This is to ensure the theme is applied properly, but may have false positives in the case of overridden packages for example
+      ''
+      // {
+        default = true;
+      };
   };
 
   config = mkIf enable {
+    assertions = lib.optional cfg.assertQt6Sddm {
+      assertion = config.services.displayManager.sddm.package == pkgs.kdePackages.sddm;
+      message = ''
+        Only the Qt 6 version of SDDM is supported by this port!
+
+        In most cases this can be resolved by setting `services.displayManager.sddm.package`
+        to `pkgs.kdePackages.sddm`. If you know what you're doing and wish to disable this check,
+        please set `services.displayManager.sddm.catppuccin.assertQt6Sddm` to `false`
+      '';
+    };
+
     services.displayManager.sddm.theme = "catppuccin-${cfg.flavor}";
+
     environment.systemPackages = [
       (pkgs.catppuccin-sddm.override {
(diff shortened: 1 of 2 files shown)


isabelroses:
closes https://github.com/catppuccin/nix/issues/229

getchoo (changes requested):
i think this would be better as an assertion...but i'm not really sure how to reliably do that as sddm doesn't directly report its qt's major version

my first thoughts are asserting that pkgs.kdePackages.qtbase is in the buildInputs of config.services.displayManager.sddm.package or making a conditionally applied assertion for services.displayManager.sddm.package via a new option

isabelroses:
I think the easiest thing would be to check what package is being used since it's either the qt5 one or the kde package.

isabelroses:
or perhaps `lib.mkDefault` would be better so we don't get clashes with users that apply the plasma 6 desktop.

getchoo:
> I think the easiest thing would be to check what package is being used since it's either the qt5 one or the kde package.

that's what i meant with a conditional assertion. it wouldn't account for overriden packages ofc, but if someone really knows what they're doing they can disable it with an option

> or perhaps lib.mkDefault would be better

i don't think it would as users could easily set the qt5 package as normal with no warning. assertions on the other hand provide a built in explanation of why we need to use the qt6 version when this happens

> so we don't get clashes with users that apply the plasma 6 desktop

why would this happen?

isabelroses:
> why would this happen?

The duplicate package deceleration I mean.

modules/nixos/sddm.nix
```diff
@@ -55,5 +56,12 @@ in
           ;
       })
     ];
+
+    assertions = [
+      {
+        assertion = config.services.displayManager.sddm.package != pkgs.kdePackages.sddm;
+        message = "Not using the kdePackages.sddm package, will cause the theme to fail being applied";
+      }
+    ];
```
getchoo, line 65:
```suggestion
    assertions = lib.optional cfg.assertQt6Sddm {
      assertion = config.services.displayManager.sddm.package != pkgs.kdePackages.sddm;
      message = "Not using the kdePackages.sddm package, will cause the theme to fail being applied";
    };
```

`assertQt6Sddm` should probably be
```nix
{
  assertQt6Sddm = lib.mkEnableOption "checking `services.displayManager.sddm.package` is the Qt 6 version. This is to ensure the theme is applied properly, but may have false positives in the case of overridden packages for example" // { default = true; };
}
```
(hopefully with better formatting :trollface:)

getchoo (changes requested):

modules/nixos/sddm.nix
isabelroses: wow that's cursed but sure.

getchoo (approved):
thanks!

PR #239, opened 2024-06-22
fix(home-manager): use correct theme name

eljamm:
## Summary
The theme name does not correctly match the one from catppuccin-gtk which should be `catppuccin-${FLAVOR}-${ACCENT}-${SIZE}+${TWEAKS}` as the [Usage docs](https://github.com/catppuccin/gtk/blob/23b52b5b9cde1e11c07315e79a55804d2ac77e3a/docs/USAGE.md#manual-installation) specify.

## Changes
- Name is all lower case now
- Order changed from `flavor-size-accent-gtkTheme` to `flavor-accent-size+tweaks`

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -72,15 +72,10 @@ in
     gtk = {
       theme =
         let
-          flavorUpper = ctp.mkUpper cfg.flavor;
-          accentUpper = ctp.mkUpper cfg.accent;
-          sizeUpper = ctp.mkUpper cfg.size;
-
-          # use the light gtk theme for latte
-          gtkTheme = if cfg.flavor == "latte" then "Light" else "Dark";
+          gtkTweaks = lib.concatStringsSep "," cfg.tweaks;
         in
         {
-          name = "Catppuccin-${flavorUpper}-${sizeUpper}-${accentUpper}-${gtkTheme}";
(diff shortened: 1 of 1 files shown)


modules/home-manager/gtk.nix
```diff
@@ -70,23 +70,14 @@ in
...
-        in
-        {
-          name = "Catppuccin-${flavorUpper}-${sizeUpper}-${accentUpper}-${gtkTheme}";
-          package = pkgs.catppuccin-gtk.override {
-            inherit (cfg) size tweaks;
-            accents = [ cfg.accent ];
-            variant = cfg.flavor;
-          };
+      theme = {
+        name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}+default";
```
isabelroses, line 74: i believe tweaks are added at the end. For example `+default` for no tweaks, `+black` for the black tweak.

eljamm: So the tweak to path mapping is this?
- black -> +black
- rimless -> +rimless
- normal -> +default

isabelroses: seem correct

eljamm:
Well, turns out I did mess up, but it should work now.

eljamm:
So apparently it's just enough to append the tweaks at the end. In the output the names were being seperated by `,` and not `+` so I changed that as well.

Screenshot from 2024-06-22 15-27-32

I tested this on my machine and everything works great.

getchoo (approved):
thanks!

PR #241, opened 2024-06-23
feat(home-manager): allow enabling GTK, cursors and icons separately

PerchunPak:
GTK port is deprecated and will be soon removed (see https://github.com/catppuccin/gtk/issues/262). And I wanted a way to use cattpuccin cursors without risking breaking my configuration during update in the near future.

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -12,7 +12,7 @@ let
     types
     ;
   cfg = config.gtk.catppuccin;
-  enable = cfg.enable && config.gtk.enable;
+  gtkEnable = cfg.enable && config.gtk.enable;
   # "dark" and "light" can be used alongside the regular accents
   cursorAccentType = ctp.mergeEnums ctp.types.accentOption (
     lib.types.enum [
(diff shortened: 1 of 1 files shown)


getchoo:
part of this is actually being done in #195 already - namely separating the cursors from the main gtk module to avoid this warning. icons are *sort of* separated in this as well, but i did plan on making a follow up PR to separate them in a similar way to the cursors. if you'd like, you could re-use some of the work here to a new `catppuccin.gtk.iconTheme` option similar to [this](https://github.com/catppuccin/nix/blob/392aff9c13abfe77c504f2c5073c8c51cc0596fe/modules/home-manager/cursor.nix) in a new PR

in any case, thank you so much for this work! as it seems we came to basically the same conclusion, i know it probably took you a good bit. sorry i have to close this :(

PerchunPak:
> i know it probably took you a good bit. sorry i have to close this :(

Don't worry, this PR took me literally 10 minutes, it is not a problem at all

I don't like Papirus icons, so don't worry about doing duplicate work again ;)

PR #242, opened 2024-06-25
fix(home-manager): assert `qt.style.name` for kvantum theme

getchoo:
while having `qt.enable` be a prerequisite for this option applying
makes sense, the requirement on a specific `style.name` is not. this
make it more clear by moving that check to an assertion rather than an
internal comparison

### DIFF
--- a/modules/home-manager/kvantum.nix
+++ b/modules/home-manager/kvantum.nix
@@ -6,13 +6,8 @@
 }:
 let
   cfg = config.qt.style.catppuccin;
-  enable =
-    cfg.enable
-    && config.qt.enable
-    && lib.elem config.qt.style.name [
-      "kvantum"
-      "Kvantum"
-    ];
+  enable = cfg.enable && config.qt.enable;
+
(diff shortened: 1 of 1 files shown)


vdbe:
Is there a reason enabling`qt.style.catppuccin` does not set `qt.style.name` to `kvantum` itself.

If so I feel like this should be a warning.

getchoo:
> Is there a reason enabling qt.style.catppuccin does not set qt.style.name to kvantum itself.

i feel some options would be better set for the user - with this being one of them. setting `qt.style.catppuccin` doesn't really give the user any hint that kvantum will be enabled, which can easily lead to issues where they may set it to something, then have a conflict to deal with that contains no real guidance on why it was set to kvantum

> If so I feel like this should be a warning.

i don't feel a warning is appropriate here as it's a hard requirement. if we can know for sure something is broken (like in the case of kvantum not being used), we should throw an evaluation error as to make it obvious `qt.style.catppuccin` will not work if enabled. after all, why would a user enable it if they were okay with it not doing anything?

PR #244, opened 2024-06-25
fix(home-manager): assert `qt.platformTheme.name` for kvantum

pinkcreeper100:


modules/home-manager/kvantum.nix
```diff
@@ -39,6 +39,10 @@ in
         ];
         message = ''`qt.style.name` must be `"kvantum"` to use `qt.style.catppuccin`'';
       }
+      {
+        assertion = lib.elem config.qt.platformTheme.name [ "kvantum" ];
```
getchoo, line 43: looks like `name` was only added in master, not release-24.05. see the error [here](https://github.com/catppuccin/nix/actions/runs/9662132054/job/26723660197?pr=244#step:5:349)
```suggestion
        assertion = lib.elem (config.qt.platformTheme.name or config.qt.platformTheme or null) [ "kvantum" ];
```

getchoo (changes requested):

getchoo (approved):
thanks!

PR #245, opened 2024-06-25
chore: remove unused argument in mk-site.nix

pinkcreeper100:


.sources/default.nix
getchoo, line 1: this should be excluded, since as the comment at the top of the file states, it will be overwritten anyways

getchoo (changes requested):
thanks btw :)

getchoo (approved):
thanks!

PR #248, opened 2024-06-27
fix(home-manager): only apply pointer cursor on linux

XYenon:
Pointer cursor is only supported on linux.

### DIFF
--- a/modules/home-manager/cursor.nix
+++ b/modules/home-manager/cursor.nix
@@ -7,6 +7,7 @@
 let
   inherit (lib) ctp mkIf;
   cfg = config.catppuccin.pointerCursor;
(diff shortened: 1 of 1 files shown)


getchoo (approved):
i've changed this to only exclude `catppuccin.pointerCursor` from the global `catppuccin.enable`, as allowing this to be set but have 0 effect is a bit odd for darwin users. the upstream [assertion](https://github.com/nix-community/home-manager/blob/7e68e55d2e16d3a1e92a679430728c35a30fd24e/modules/config/home-cursor.nix#L134) should take care of the rest in case someone sets it manually

thanks for catching this btw! hopefully with darwin testing now, this won't happen again so easily :p

PR #261, opened 2024-06-30
fix(home-manager/gtk): replace `normal` tweak with `default`

isabelroses:
Seeing as this does not exist and should be default we should remove the "normal" and replace it with default
image

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -121,7 +121,9 @@ in
     (mkIf enable {
       gtk.theme =
         let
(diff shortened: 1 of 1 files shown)


jpotier:
I'm quite confused as I'm facing the opposite problem now: default doesn't exist, but normal does…

image

jpotier:
This makes no sense:

image

Upstream https://github.com/catppuccin/gtk/releases clearly has "default" suffix.

isabelroses:
Yeahhhh... we had this exact same problem right after fixing it too. The issue is I keep getting weird builds with default and others with normal. I think we should just revert this @getchoo .

PR #274, opened 2024-07-03
fix(home-manager/gtk): support all tweaks

isabelroses:
gtk will be the death of me

### DIFF
--- a/modules/home-manager/gtk.nix
+++ b/modules/home-manager/gtk.nix
@@ -42,9 +42,10 @@ in
             "black"
             "rimless"
             "normal"
+            "float"
           ]
         );
-        default = [ "normal" ];
+        default = [ ];
         description = "Catppuccin tweaks for gtk";
       };
 
@@ -121,10 +122,12 @@ in
     (mkIf enable {
       gtk.theme =
         let
-          gtkTweaks = concatStringsSep "," cfg.tweaks;
+          gtkTweaks = "+" + concatStringsSep "," cfg.tweaks;
         in
         {
-          name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}+${gtkTweaks}";
+          name =
+            "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}"
+            + lib.optionalString (cfg.tweaks != [ ]) gtkTweaks;
           package = pkgs.catppuccin-gtk.override {
             inherit (cfg) size tweaks;
             accents = [ cfg.accent ];


isabelroses:
Will need changes after https://github.com/NixOS/nixpkgs/pull/324265. Worth waiting till its merged and backported.

modules/home-manager/gtk.nix
```diff
@@ -121,7 +122,7 @@ in
     (mkIf enable {
       gtk.theme =
         let
-          gtkTweaks = concatStringsSep "," cfg.tweaks;
+          gtkTweaks = if cfg.tweaks == [ ] then "default" else (concatStringsSep "," cfg.tweaks);
         in
         {
           name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}+${gtkTweaks}";
```
eljamm, line 128: Following the changes from the nixpkgs PR, if tweaks is an empty list, the name will become:

```
"catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}"
```

In which case we'd need to do something like this:

```suggestion
          gtkTweaks = if cfg.tweaks == [ ] then "" else (concatStrings [ "+" (concatStringsSep "," cfg.tweaks) ]);
        in
        {
          name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}${gtkTweaks}";
```

modules/home-manager/gtk.nix
```diff
@@ -121,10 +122,10 @@ in
     (mkIf enable {
       gtk.theme =
         let
-          gtkTweaks = concatStringsSep "," cfg.tweaks;
+          gtkTweaks = if cfg.tweaks == [ ] then "" else (concatStrings [ "+" (concatStringsSep "," cfg.tweaks) ]);
```
eljamm, line 125: Apologies, I didn't format the suggestion...

```suggestion
          gtkTweaks =
            if cfg.tweaks == [ ] then
              ""
            else
              (concatStrings [
                "+"
                (concatStringsSep "," cfg.tweaks)
              ]);
```

isabelroses:
Both PRs hit their respective branches, so we should be good to merge now.

eljamm (approved):

modules/home-manager/gtk.nix
```diff
@@ -121,10 +123,17 @@ in
...
         let
-          gtkTweaks = concatStringsSep "," cfg.tweaks;
+          gtkTweaks =
+            if cfg.tweaks == [ ] then
+              ""
+            else
+              (concatStrings [
+                "+"
+                (concatStringsSep "," cfg.tweaks)
+              ]);
```
getchoo, line 133:
```suggestion
          gtkTweaks = "+" + concatStringsSep "," cfg.tweaks;
```
there are like a thousand different ways you could do this, but i just found this to be a bit cleaner

modules/home-manager/gtk.nix
```diff
@@ -121,10 +123,17 @@ in
...
         in
         {
-          name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}+${gtkTweaks}";
+          name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}${gtkTweaks}";
```
getchoo, line 136:
```suggestion
          name = "catppuccin-${cfg.flavor}-${cfg.accent}-${cfg.size}" + lib.optionalString (cfg.tweaks != [ ]) gtkTweaks;
```

modules/home-manager/gtk.nix
```diff
@@ -6,6 +6,7 @@
 }:
 let
   inherit (lib)
+    concatStrings
```
getchoo, line 9:
```suggestion
```

getchoo (changes requested):

modules/home-manager/gtk.nix
eljamm: Indeed, that's much better.

getchoo (approved):
thanks! hopefully this will be the last gtk PR now...

PR #276, opened 2024-07-04
fix(home-manager/cursors): automatically enable cursor for gtk and x11

PerchunPak:
Otherwise, `home.pointerCursor` is useless I suppose

EDIT: Not entirely useless (see https://github.com/nix-community/home-manager/blob/0a30138c694ab3b048ac300794c2eb599dc40266/modules/config/home-cursor.nix#L165), but still it didn't work for me on GNOME

### DIFF
--- a/modules/home-manager/cursor.nix
+++ b/modules/home-manager/cursor.nix
@@ -31,5 +31,8 @@ in
   config.home.pointerCursor = mkIf cfg.enable {
     name = "catppuccin-${cfg.flavor}-${cfg.accent}-cursors";
     package = pkgs.catppuccin-cursors.${cfg.flavor + ctp.mkUpper cfg.accent};
+
+    gtk.enable = true;
+    x11.enable = true;
   };
 }


isabelroses:
Personally I think this would be better as an assertion, it its added at all. Perhaps this should assert that at least one of these are set to true?

Weathercold:
An assertion doesn’t make sense because `home.pointerCursor` is effective by itself, and some people might want to leave the extra options disabled. I suggest we either don’t set them or set them with `mkDefault`

getchoo:
unlike some other places where we assert or enable options, this is simply changing the defaults of the parent module that we don't actually require. suggestions like this would probably be better made in the upstream module, as this is not specific to us and i would consider this intended behavior at it stands. thanks though :)

PerchunPak:
Maybe we should add a note/warning in docs to let users know that they should enable those options in some cases in order to make cursor work?

PR #296, opened 2024-07-28
feat(home-manager): add support for spotify-player

w-lfchen:
I have an issue with the theme:
The `playback_metadata` component has the same color as the background and is therefore not legible.
But this should be fixed in the [catppuccin repo](https://github.com/catppuccin/spotify-player), right?

### DIFF
--- a/modules/home-manager/spotify-player.nix
+++ b/modules/home-manager/spotify-player.nix
@@ -1,0 +1,16 @@
+{ config, lib, ... }:
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.programs.spotify-player.catppuccin;
+  enable = cfg.enable && config.programs.spotify-player.enable;
+in
+{
+  options.programs.spotify-player.catppuccin = lib.ctp.mkCatppuccinOpt { name = "spotify-player"; };
+
+  config = lib.mkIf enable {
+    programs.spotify-player = {
+      settings.theme = "Catppuccin-${cfg.flavor}";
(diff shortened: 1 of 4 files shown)


modules/home-manager/spotify-player.nix
```diff
@@ -0,0 +1,16 @@
...
+  cfg = config.programs.spotify-player.catppuccin;
+  enable = cfg.enable && config.programs.spotify-player.enable;
+in
+{
+  options.programs.spotify-player.catppuccin = lib.ctp.mkCatppuccinOpt { name = "spotify-player"; };
+
+  config = lib.mkIf enable {
+    programs.spotify-player = {
+      settings.theme = "Catppuccin-${cfg.flavor}";
+      themes = (lib.importTOML "${sources.spotify-player}/theme.toml").themes;
```
getchoo, line 13:
```suggestion
      inherit (lib.importTOML "${sources.spotify-player}/theme.toml") themes;
```

getchoo (changes requested):
This should also add a new entry to `tests/home.nix` enabling the module

> But this should be fixed in the [catppuccin repo](https://github.com/catppuccin/spotify-player), right?

And yep!

w-lfchen:
didn't mean to make the ci status disappear by updating the branch ^^'
here are the results of the run:
https://github.com/catppuccin/nix/actions/runs/10280254042

getchoo (approved):
Thanks!

PR #299, opened 2024-07-30
fix(home-manager/hyprland): inherit cursor size, unset hyprcursor env vars

### LINKED ISSUE #236: Hyprcursor theme is set to "MyCursor"
https://github.com/catppuccin/nix/blob/73e06d5bd7ed34bdd0168030893ef8364fdc1d4a/modules/home-manager/hyprland.nix#L30

The hyprland module sets up environmental variables for the cursor, however these are overwritten later where the [example code from the hyprland wiki](https://wiki.hyprland.org/Hypr-Ecosystem/hyprcursor/) is added to the hyprland config, setting the cursor theme to `MyCursor` instead of `catppuccin-${cursor.flavor}-${cursor.accent}-cursors`

Weathercold:
Fixes #236

### DIFF
--- a/modules/home-manager/hyprland.nix
+++ b/modules/home-manager/hyprland.nix
@@ -18,20 +18,13 @@ in
     };
 
     wayland.windowManager.hyprland.settings = {
-      source =
-        [
-          "${sources.hyprland}/themes/${cfg.flavor}.conf"
-          (builtins.toFile "hyprland-${cfg.accent}-accent.conf" ''
-            $accent=''$${cfg.accent}
-            $accentAlpha=''$${cfg.accent}Alpha
-          '')
-        ]
-        ++ lib.optionals pointerCursor.enable [
-          (builtins.toFile "hyprland-cursors.conf" ''
-            env = HYPRCURSOR_THEME,MyCursor
-            env = HYPRCURSOR_SIZE,24
-          '')
-        ];
+      source = [
+        "${sources.hyprland}/themes/${cfg.flavor}.conf"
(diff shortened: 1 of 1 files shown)


getchoo (approved):
Thanks!

modules/home-manager/hyprland.nix
```diff
@@ -14,25 +13,14 @@ in
...
-        ]
-        ++ lib.optionals pointerCursor.enable [
-          (builtins.toFile "hyprland-cursors.conf" ''
-            env = HYPRCURSOR_THEME,MyCursor
-            env = HYPRCURSOR_SIZE,24
-          '')
-        ];
+      source = "${sources.hyprland}/themes/${cfg.flavor}.conf";
+      "$accent" = "\$${cfg.accent}";
+      "$accentAlpha" = "\$${cfg.accent}Alpha";
```
PerchunPak, line 23: These values were provided as a file to source because nix may decide to define variables of the user before accent colors. Thus, user may not be able to use those accent colors in custom variables

Example:
```nix
wayland.windowManager.hyprland.settings = {
  # inside ctp
  "$accent" = "\$${cfg.accent}";
  # user config
  "$AAAmyvar" = "<span color="$accent">...</span>";
};
```
Could result to
```ini
source=/nix/store/...-hyprland/themes/...
$AAAmyvar = <span color="$accent">...</span> # oops, $accent is an empty string
$accent = $magenta
# ...
```

Weathercold: I'll revert sometime

Weathercold: #347

PR #300, opened 2024-08-05
fix(home-manager/foot): avoid IFD

Anomalocaridid:
Use foot's `include` option to import the theme rather than importing the theme into nix.


getchoo (approved):

getchoo:
Congrats on making the 300th PR! 🥳

PR #302, opened 2024-08-05
fix(home-manager/dunst): avoid IFD

Anomalocaridid:
Uses dunst's drop-in feature to set the theme rather than importing it into nix.

This approach has the drawback that theme settings are now potentially slightly more inconvenient to override. The theme drop-in overrides settings from `dunstrc`, but it can itself be overridden by another drop-in that is after it in lexical order.

I added a `prefix` option to make it more convenient to have the catppuccin drop-in appear in a certain spot in lexical order if the user has multiple drop-ins they wish to use with dunst.

Currently blocked on https://github.com/nix-community/home-manager/pull/5710

### DIFF
--- a/modules/home-manager/dunst.nix
+++ b/modules/home-manager/dunst.nix
@@ -5,9 +5,22 @@ let
   enable = cfg.enable && config.services.dunst.enable;
 in
 {
-  options.services.dunst.catppuccin = lib.ctp.mkCatppuccinOpt { name = "dunst"; };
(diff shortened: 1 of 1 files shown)


Anomalocaridid:
https://github.com/nix-community/home-manager/pull/5710 has been merged, so this is no longer blocked.

getchoo (approved):
LGTM, but I'm going to leave this open for a bit as I don't use dunst personally and would like to give others some time to give their opinion on this. Thanks!

Anomalocaridid:
Fair enough. Better safe than sorry.

PR #304, opened 2024-08-06
fix(home-manager/lazygit): avoid IFD

Anomalocaridid:
Sets `$LG_CONFIG_FILE` to merge the default lazygit config file and the theme file rather than importing the theme file into nix.

### DIFF
--- a/modules/home-manager/lazygit.nix
+++ b/modules/home-manager/lazygit.nix
@@ -4,15 +4,14 @@ let
   inherit (config.catppuccin) sources;
   cfg = config.programs.lazygit.catppuccin;
   enable = cfg.enable && config.programs.lazygit.enable;
-
-  themePath = "${cfg.flavor}/${cfg.accent}.yml";
 in
 {
   options.programs.lazygit.catppuccin = lib.ctp.mkCatppuccinOpt { name = "lazygit"; } // {
     accent = ctp.mkAccentOpt "lazygit";
   };
 
-  config = lib.mkIf enable {
-    programs.lazygit.settings = lib.ctp.fromYaml "${sources.lazygit}/themes-mergable/${themePath}";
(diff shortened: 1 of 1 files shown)


modules/home-manager/lazygit.nix
```diff
@@ -4,15 +4,14 @@ let
...
 {
   options.programs.lazygit.catppuccin = lib.ctp.mkCatppuccinOpt { name = "lazygit"; } // {
     accent = ctp.mkAccentOpt "lazygit";
   };
 
-  config = lib.mkIf enable {
-    programs.lazygit.settings = lib.ctp.fromYaml "${sources.lazygit}/themes-mergable/${themePath}";
+  config.home.sessionVariables = lib.mkIf enable {
+    # Ensure that the default config file is still sourced
+    LG_CONFIG_FILE = "${config.xdg.configHome}/lazygit/config.yml,${sources.lazygit}/themes-mergable/${cfg.flavor}/${cfg.accent}.yml";
```
getchoo, line 15: From a quick read of the [docs](https://github.com/jesseduffield/lazygit/blob/a3560eb451fcf375378a316cbc2f4e08e24bf756/docs/Config.md#overriding-default-config-file-location) it doesn't seem to be explicitly stated which file here would take precedence in the case of conflicting options. Based on the example of `LG_CONFIG_FILE="$HOME/.base_lg_conf,$HOME/.light_theme_lg_conf"` though, it seems to be latter; if this is the case I would much rather have this changed to have the user config file take precedence over our own

getchoo (changes requested):

modules/home-manager/lazygit.nix
Anomalocaridid: Fixed. Also, I tested it and I can confirm that the latter entry would take precedence.

getchoo (approved):
thanks!

Issue #308, opened 2024-08-09
bug: `hyprland.settings.source` does not merge list


zspher:
I source multiple files via `hyprland.settings.source`.
The ff. code does not allow multiple hyprland sources

https://github.com/catppuccin/nix/blob/66f4ea170093b62f319f41cebd2337a51b225c5a/modules/home-manager/hyprland.nix#L21

pls change to 

```nix
source = ["${sources.hyprland}/themes/${cfg.flavor}.conf"];
```

PR #309, opened 2024-08-09
fix(home-manager/hyprland): allow merging `sources` option

zspher:
fixes https://github.com/catppuccin/nix/issues/308


getchoo (approved):
thanks!

Issue #310, opened 2024-08-12
Path of k9s skin is broken on Darwin when XDG_CONFIG_HOME is unset


m-wynn:
On Darwin, If `XDG_CONFIG_HOME` is set, k9s will use that path for its files (e.g. `~/.config/k9s`.  If it is unset, k9s will use `~/Library/Application\ Support/k9s`.

Right now, catppuccin/nix will always use the [xdg path](https://github.com/catppuccin/nix/blob/main/modules/home-manager/k9s.nix#L19).

[Home-manager](https://github.com/nix-community/home-manager/blob/086f619dd991a4d355c07837448244029fc2d9ab/modules/programs/k9s.nix#L172-L180) gets around this with the following:

```nix
    skinFiles = mapAttrs' (name: value:
      nameValuePair (if !(isDarwin && !config.xdg.enable) then
        "k9s/skins/${name}.yaml"
      else
        "Library/Application Support/k9s/skins/${name}.yaml") {
          source = yamlFormat.generate "k9s-skin-${name}.yaml" value;
        }) cfg.skins;
```

Of course, an easy workaround is to enable XDG in your home-manager config and then export the XDG environment variables:
```nix
xdg = {
  enable = true;
};
```

alejandro-angulo:
I ran into this exact issue earlier today. I opened up PR #311 .

@m-wynn Would you mind testing out the change if you get a chance please? My repo is https://github.com/alejandro-angulo/catppuccin-nix/.

m-wynn:
Thanks @alejandro-angulo.  This worked for me as expected both with and without xdg.

PR #311, opened 2024-08-13
fix(home-manager/k9s): support darwin without XDG

alejandro-angulo:
k9s expects configuration files on OS X to live in `~/Library/Application Support/k9s` instead of `~/.config.k9s`. This commit handles the case where the module is built for a darwin system with xdg disabled.

See: https://github.com/catppuccin/nix/issues/310

### DIFF
--- a/modules/home-manager/k9s.nix
+++ b/modules/home-manager/k9s.nix
@@ -1,22 +1,32 @@
-{ config, lib, ... }:
-let
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
   inherit (config.catppuccin) sources;
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
 
   cfg = config.programs.k9s.catppuccin;
   enable = cfg.enable && config.programs.k9s.enable;
 
   themeName = "catppuccin-${cfg.flavor}" + lib.optionalString cfg.transparent "-transparent";
   themeFile = "${themeName}.yaml";
-  themePath = "/skins/${themeFile}";
+  configDirectory =
+    if (isDarwin && !config.xdg.enable)
+    then "/Library/Application Support/k9s"
+    else "k9s";
+  themePath = "${configDirectory}/skins/${themeFile}";
   theme = sources.k9s + "/dist/${themeFile}";
-in
-{
-  options.programs.k9s.catppuccin = lib.ctp.mkCatppuccinOpt { name = "k9s"; } // {
-    transparent = lib.mkEnableOption "transparent version of flavor";
-  };
+in {
+  options.programs.k9s.catppuccin =
+    lib.ctp.mkCatppuccinOpt {name = "k9s";}
+    // {
+      transparent = lib.mkEnableOption "transparent version of flavor";
+    };
 
   config = lib.mkIf enable {
-    xdg.configFile."k9s${themePath}".source = theme;
+    xdg.configFile."${themePath}".source = theme;
 
     programs.k9s.settings.k9s.ui.skin = themeName;
   };
--- a/modules/home-manager/k9s.nix
+++ b/modules/home-manager/k9s.nix
@@ -3,31 +3,32 @@
   lib,
   pkgs,
   ...
-}: let
+}:
+let
   inherit (config.catppuccin) sources;
-  inherit (pkgs.stdenv.hostPlatform) isDarwin;
 
   cfg = config.programs.k9s.catppuccin;
   enable = cfg.enable && config.programs.k9s.enable;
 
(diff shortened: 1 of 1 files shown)


alejandro-angulo:
I ran into a similar issue with lazygit (see #312 ). Would it be easier to maintain this flake if there were an assertion ensuring XDG is enabled rather than sprinkling in darwin checks?

m-wynn:
This worked for me as expected on Darwin both with and without xdg.

modules/home-manager/k9s.nix
```diff
@@ -1,22 +1,32 @@
-{ config, lib, ... }:
-let
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
   inherit (config.catppuccin) sources;
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
```
getchoo, line 8:
```suggestion
```

modules/home-manager/k9s.nix
```diff
@@ -1,22 +1,32 @@
...
 
   cfg = config.programs.k9s.catppuccin;
   enable = cfg.enable && config.programs.k9s.enable;
```
getchoo, line 11:
```suggestion
  enable = cfg.enable && config.programs.k9s.enable;
  
  # NOTE: On MacOS specifically, k9s expects it's configuration to be in
  # `~/Library/Application Support` when not using XDG
  enableXdgConfig = !pkgs.stdenv.hostPlatform.isDarwin || config.xdg.enable;
```

modules/home-manager/k9s.nix
```diff
@@ -1,22 +1,32 @@
...
 
   themeName = "catppuccin-${cfg.flavor}" + lib.optionalString cfg.transparent "-transparent";
   themeFile = "${themeName}.yaml";
-  themePath = "/skins/${themeFile}";
+  configDirectory =
+    if (isDarwin && !config.xdg.enable)
+    then "/Library/Application Support/k9s"
+    else "k9s";
+  themePath = "${configDirectory}/skins/${themeFile}";
```
getchoo, line 19:
```suggestion
  themePath = "k9s/skins/${themeFile}";
```

modules/home-manager/k9s.nix
```diff
@@ -1,22 +1,32 @@
...
+    // {
+      transparent = lib.mkEnableOption "transparent version of flavor";
+    };
 
   config = lib.mkIf enable {
-    xdg.configFile."k9s${themePath}".source = theme;
+    xdg.configFile."${themePath}".source = theme;
 
     programs.k9s.settings.k9s.ui.skin = themeName;
   };
```
getchoo, line 32:
```nix
  config = lib.mkIf enable (
    lib.mkMerge [
      (lib.mkIf (!enableXdgConfig) {
        home.file."Library/Application Support/${themePath}".source = theme;
      })
      (lib.mkIf enableXdgConfig { xdg.configFile.${themePath}.source = theme; })

      { programs.k9s.settings.k9s.ui.skin = themeName; }
    ]
  );
```

getchoo (changes requested):
It would probably be better to follow the [upstream module](https://github.com/nix-community/home-manager/blob/2598861031b78aadb4da7269df7ca9ddfc3e1671/modules/programs/k9s.nix#L180)'s example here

This should also be formatted with `nixfmt-rfc-style`

modules/home-manager/k9s.nix
```diff
@@ -1,23 +1,34 @@
...
   options.programs.k9s.catppuccin = lib.ctp.mkCatppuccinOpt { name = "k9s"; } // {
     transparent = lib.mkEnableOption "transparent version of flavor";
   };
 
-  config = lib.mkIf enable {
-    xdg.configFile."k9s${themePath}".source = theme;
-
-    programs.k9s.settings.k9s.ui.skin = themeName;
-  };
+  config = lib.mkIf enable lib.mkMerge [
```
getchoo, line 27:
```suggestion
  config = lib.mkIf enable (lib.mkMerge [
```

modules/home-manager/k9s.nix
```diff
@@ -1,23 +1,34 @@
...
+    (lib.mkIf (!enableXdgConfig) {
+      home.file."Library/Application Support/${themePath}".source = theme;
+    })
+    (lib.mkIf enableXdgConfig { xdg.configFile.${themePath}.source = theme; })
+    { programs.k9s.settings.k9s.ui.skin = themeName; }
+  ];
```
getchoo, line 33:
```suggestion
  ]);
```

getchoo (changes requested):

getchoo (approved):
Thanks!

PR #313, opened 2024-08-14
fix(home-manager/lazygit): support darwin without XDG

### LINKED ISSUE #312: lazygit does not launch on OS X when XDG is not enabled
Lazygit fails to start after enabling the catppuccin integration

```
❯ lazygit
2024/08/13 10:13:45 stat /Users/alejandro/.config/lazygit/config.yml: no such file or directory
```

It works with xdg enabled though. I can get it to launch if I unset the `LG_CONFIG_FILE` variable. I think this might've been introduced with in commit 66f4ea170093b62f319f41cebd2337a51b225c5a . We should probably check to make sure xdg is enabled before constructing paths with with `configHome`.

Problem seems similar to #310 .

alejandro-angulo:
Lazygit would fail to load becase the wrong path was being generated to the configuration file on darwin systems when XDG wasn't enabled.

Fixes #312

### DIFF
--- a/modules/home-manager/lazygit.nix
+++ b/modules/home-manager/lazygit.nix
@@ -1,17 +1,28 @@
-{ config, lib, ... }:
-let
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
   inherit (lib) ctp;
   inherit (config.catppuccin) sources;
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
+
   cfg = config.programs.lazygit.catppuccin;
   enable = cfg.enable && config.programs.lazygit.enable;
-in
-{
-  options.programs.lazygit.catppuccin = lib.ctp.mkCatppuccinOpt { name = "lazygit"; } // {
-    accent = ctp.mkAccentOpt "lazygit";
-  };
+  configDirectory =
+    if (isDarwin && !config.xdg.enable)
+    then "${config.home.homeDirectory}/Library/Application Support/lazygit"
+    else "${config.xdg.configHome}/lazygit";
+in {
+  options.programs.lazygit.catppuccin =
+    lib.ctp.mkCatppuccinOpt {name = "lazygit";}
+    // {
+      accent = ctp.mkAccentOpt "lazygit";
+    };
 
   config.home.sessionVariables = lib.mkIf enable {
     # Ensure that the default config file is still sourced
-    LG_CONFIG_FILE = "${sources.lazygit}/themes-mergable/${cfg.flavor}/${cfg.accent}.yml,${config.xdg.configHome}/lazygit/config.yml";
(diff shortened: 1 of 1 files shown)


modules/home-manager/lazygit.nix
```diff
@@ -1,17 +1,28 @@
...
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}: let
   inherit (lib) ctp;
   inherit (config.catppuccin) sources;
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
+
```
getchoo, line 10:
```suggestion

  # NOTE: On MacOS specifically, lazygit expects it's configuration to be in
  # `~/Library/Application Support` when not using XDG
  enableXdgConfig = !pkgs.stdenv.hostPlatform.isDarwin || config.xdg.enable;
```

modules/home-manager/lazygit.nix
```diff
@@ -1,17 +1,28 @@
...
   enable = cfg.enable && config.programs.lazygit.enable;
-in
-{
-  options.programs.lazygit.catppuccin = lib.ctp.mkCatppuccinOpt { name = "lazygit"; } // {
-    accent = ctp.mkAccentOpt "lazygit";
-  };
+  configDirectory =
+    if (isDarwin && !config.xdg.enable)
+    then "${config.home.homeDirectory}/Library/Application Support/lazygit"
+    else "${config.xdg.configHome}/lazygit";
```
getchoo, line 16:
```suggestion
  configDirectory =
    if (enableXdgConfig)
    then "${config.home.homeDirectory}/Library/Application Support"
    else config.xdg.configHome;
  configFile = "${configDirectory}/lazygit/config.yml";
```

modules/home-manager/lazygit.nix
```diff
@@ -1,17 +1,28 @@
...
+  options.programs.lazygit.catppuccin =
+    lib.ctp.mkCatppuccinOpt {name = "lazygit";}
+    // {
+      accent = ctp.mkAccentOpt "lazygit";
+    };
 
   config.home.sessionVariables = lib.mkIf enable {
     # Ensure that the default config file is still sourced
-    LG_CONFIG_FILE = "${sources.lazygit}/themes-mergable/${cfg.flavor}/${cfg.accent}.yml,${config.xdg.configHome}/lazygit/config.yml";
+    LG_CONFIG_FILE = "${sources.lazygit}/themes-mergable/${cfg.flavor}/${cfg.accent}.yml,${configDirectory}/config.yml";
```
getchoo, line 26:
```suggestion
    LG_CONFIG_FILE = "${sources.lazygit}/themes-mergable/${cfg.flavor}/${cfg.accent}.yml,${configFile}";
```

getchoo (changes requested):

getchoo (approved):
thanks!

PR #322, opened 2024-08-30
fix: add enable option to fuzzel

SchweGELBin:
It only checked if catppuccin was enabled globally


getchoo:
Thanks for the patch! This is actually intentional though, as the upstream module [already has this guard](https://github.com/nix-community/home-manager/blob/471e3eb0a114265bcd62d11d58ba8d3421ee68eb/modules/programs/fuzzel.nix#L39)

SchweGELBin:
Oh, thank you. I didn't know about it, just wondered, why this is the only module that did it differently.
Thank you for the explaination and have a great day!

PR #327, opened 2024-08-31
feat(home-manager): add support for freetube

Anomalocaridid:
Adds support for theming FreeTube. The colors are built-in, so no new sources have been added.

Also, FreeTube supports two different accent colors, so I added a `secondaryAccent` option. I made it default to FreeTube's `accent` rather than the global `accent` setting because it seemed like it would be intuitive behavior to have them match if, say, only `accent` is overridden.

### DIFF
--- a/modules/home-manager/freetube.nix
+++ b/modules/home-manager/freetube.nix
@@ -1,0 +1,25 @@
+{ config, lib, ... }:
+let
+  inherit (config.programs.freetube.settings) baseTheme;
+  inherit (lib.ctp) mkAccentOpt mkUpper;
+  cfg = config.programs.freetube.catppuccin;
+  enable = cfg.enable && config.programs.freetube.enable;
+in
+{
+  options.programs.freetube.catppuccin = lib.ctp.mkCatppuccinOpt { name = "freetube"; } // {
+    accent = mkAccentOpt "FreeTube";
+    # FreeTube supports two accent colors
+    secondaryAccent = mkAccentOpt "FreeTube" // {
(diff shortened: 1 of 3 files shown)


modules/home-manager/freetube.nix
```diff
@@ -0,0 +1,24 @@
+{ config, lib, ... }:
+let
+  inherit (lib.ctp) mkAccentOpt mkUpper;
```
getchoo, line 3:
```suggestion
  inherit (config.programs.freetube.settings) baseTheme;
  inherit (lib.ctp) mkAccentOpt mkUpper;
```

modules/home-manager/freetube.nix
```diff
@@ -0,0 +1,24 @@
...
+    accent = mkAccentOpt "FreeTube";
+    # FreeTube supports two accent colors
+    secondaryAccent = mkAccentOpt "FreeTube" // {
+      # Have the secondary accent default to FreeTube's main accent rather than the global Catppuccin accent
+      # This assumes most users would prefer both accent colors to be the same when only overriding the main one
+      default = cfg.accent;
+    };
+  };
+
+  config.programs.freetube.settings = lib.mkIf enable rec {
```
getchoo, line 18:
```suggestion
  config.programs.freetube.settings = lib.mkIf enable {
```

getchoo (changes requested):

getchoo (approved):
LGTM. Sorry for the long wait and thanks!

Anomalocaridid:
No worries! Thank you too!

Issue #295, opened 2024-07-27
Hyprlock support


kcoderhtml:
This was originally going to be addressed in this pr #293 but the author has closed it. For a new integration I think that adding fields that can optionally set the face and background images would be the best course of action around not having a standard location set as mentioned in the above pr

I don’t have a ton of nix experience but I’d be willing to help try and get this integrated!

42Willow:
This issue can be closed now :)

PR #293, opened 2024-07-23
feat(home-manager): add support for hyprlock

PerchunPak:
https://github.com/catppuccin/hyprlock

PR #330, opened 2024-09-04
feat(home-manager): add support for hyprlock

42Willow:
Almost identical to the hyprland home manager module, links the catppuccin source file so colour variables can be used in the config. 
image

https://github.com/catppuccin/hyprlock

Thank you to @PerchunPak for her work in #293


42Willow:
Fixes #295 

Config has been tested and is working

42Willow:
Oops, just realised this doesn’t actually use the hyperlink port at all, so I need to remove that from sources.json

The hyprlock port is only a configuration file? Why does it even exist when the [hyprland](https://github.com/catppuccin/hyprland) port is all you need?

getchoo (changes requested):
This module should be enabled in `tests/home.nix`

42Willow:
> This module should be enabled in `tests/home.nix`

Should I disable the test in https://github.com/catppuccin/nix/blob/45745fe5960acaefef2b60f3455bcac6a0ca6bc9/tests/darwin.nix ?

getchoo (approved):
Sorry for the wait. Thanks!

Issue #215, opened 2024-06-07
Support integrating with plasma-manager for theming Plasma, Konsole, ...


pluiedev:
Theming KDE apps, including Plasma itself, has always been a tricky matter. Since KDE always expects its configuration files to be freely writable, the typical approach to simply make the file a symlink of a generated config file in the store is not viable. However, catppuccin-nix could integrate with [plasma-manager](https://github.com/pjones/plasma-manager) to support KDE theming, as it already has all the foundations needed to correctly handle KDE config files.

However, there is one problem with supporting plasma-manager — it has quite a hefty size, and not everyone requires using it. Since flakes as of now don't support optional dependencies, that means we could potentially end up needing a different version of the main flake that does offer plasma-manager support. Would that amount of work really be worth it to support KDE?

hyshka:
As a Plasma and Catppuccin Nix user, I'd more than welcome a separate flake if it configured the Plasma theme for me. Would that reduce the complexity significantly?

pluiedev:
> As a Plasma and Catppuccin Nix user, I'd more than welcome a separate flake if it configured the Plasma Catppuccin theme for me. Would that reduce the complexity significantly?

In my experience, not really, since I already maintain something akin to it [in my own repo](https://github.com/pluiedev/flake/tree/main/hm-modules/ctp-plus). Although, it's important that catppuccin-nix's lib module be made public in order to avoid some dirty hacks I've been using to gain access to it — I've talked to @getchoo about it and it still hasn't happened yet, though I can see why as flakes still lack versioning or any sort of stable API guarantees

PR #332, opened 2024-09-05
feat(plasma-manager): add support for konsole

HeitorAugustoLN:
Adds plasma manager modules support and Konsole support in plasma-manager module

Related: #215


HeitorAugustoLN:
Didn't found a good solution for it, so I will close it for now.

PR #337, opened 2024-09-20
fix(home-manager/kitty): use new `themeFile` option on 24.11

olifloof:
also removed kitty from the IFD section in the faq, as `themeFile` doesn't use IFD

### DIFF
--- a/modules/home-manager/kitty.nix
+++ b/modules/home-manager/kitty.nix
@@ -3,9 +3,13 @@ let
   inherit (lib) ctp;
   cfg = config.programs.kitty.catppuccin;
   enable = cfg.enable && config.programs.kitty.enable;
+
+  # TODO: Remove after 24.11 is stable
+  # https://github.com/nix-community/home-manager/pull/5750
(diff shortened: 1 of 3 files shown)


olifloof:
also note: this requires the `master` branch of home-manager

olifloof:
i should probably find a way to make it work for older versions

modules/home-manager/kitty.nix
```diff
@@ -7,5 +7,5 @@ in
 {
   options.programs.kitty.catppuccin = ctp.mkCatppuccinOpt { name = "kitty"; };
 
-  config.programs.kitty = lib.mkIf enable { theme = "Catppuccin-${ctp.mkUpper cfg.flavor}"; };
+  config.programs.kitty = lib.mkIf enable { themeFile = "Catppuccin-${ctp.mkUpper cfg.flavor}"; };
```
khaneliman, line 10: This should be a path, I'd assume? You are still just passing the name of a theme.

getchoo:
Waiting on #306 to pull in a newer revision of home-manager for CI

PR #342, opened 2024-09-22
docs: remove kitty from README ifd faq

vdbe:
Upstream removed the ifd in home-manger/kitty, #337 made the required changes and removed it from the ifd faq in docs/src/faq.md but not README.md.


getchoo (approved):
Thanks!

PR #351, opened 2024-10-12
fix: add nix darwin support

Kevin-2483:
i have use this config for month, Although there are only a few apps supported, but it just work. that is enough


getchoo (commented):
I don't think we should be making this an output without any supported ports. It makes these options (and importing the module in general) kinda pointless

Kevin-2483:
Although they are meaningless for the time being, they do not affect the build results, and may be supported in the future。Maybe I can set up some options to turn them off, but I don't want to spend too much time on things that don't affect the results anymore。so，i well close this request and use my fork in my computer。Thanks a lot for help。

PR #370, opened 2024-11-10
fix(fuzzel): correctly gatekeep module

isabelroses:


isabelroses:
build is failing on darwin due to spotify-player

getchoo (changes requested):
The [upstream module](https://github.com/nix-community/home-manager/blob/60bb110917844d354f3c18e05450606a435d2d10/modules/programs/fuzzel.nix#L39) already gates this behind `programs.fuzzel.enable`. This shouldn't be needed

isabelroses:
> The [upstream module](https://github.com/nix-community/home-manager/blob/60bb110917844d354f3c18e05450606a435d2d10/modules/programs/fuzzel.nix#L39) already gates this behind `programs.fuzzel.enable`. This shouldn't be needed

the same could be said for fish and others but we do the same thing https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/fish.nix#L5

getchoo:
> the same could be said for fish and others but we do the same thing

That was honestly out of laziness when refactoring. Some home-manager modules will (correctly) guard applying their configs when they're not enabled, while others wont. Ideally we should be checking for this and only adding this guard when we have to (since it's pretty annoying)

Issue #222, opened 2024-06-10
Namespace our module options


getchoo:
Currently we use existing option namespaces for our own (i.e., our options for `programs.bat` are under `programs.bat.catppuccin`), which can introduce some issues in options that use submodule as their type. To avoid this, it'd probably be best to use the already existing `catppuccin` namespace for all of our options (an example of this would be `catppuccin.programs.bat`)

This would break basically every option we have currently, but would hopefully avoid problems in the future as things like the freeform type become more widely adopted

Thanks to @Weathercold for suggesting this

isabelroses:
This doesn't seem like a bad Idea when you take into accound we are creating our own `programs.glamour` which might be taken by another program later if home-manager does something.

See: https://github.com/catppuccin/nix/blob/85558d1638a65ed1cc7fb8bd7cfc1a5474b21fdb/modules/home-manager/glamour.nix

Airradda:
As a user, this will make it a lot easier to quickly enable all the module options with a copy paste of the whole `catppuccin.*`, as well as easily tell which I have enabled, which will be especially important as more of them get continually added. The only issue/inconvenience I can imagine, for me at least, is in the symbol picker of a [Nil](https://github.com/oxalica/nil)+[Helix](https://github.com/helix-editor/helix/) these module options will likely be sorted before the Home-Manager ones (`catppuccin.programs.bat` then `programs.bat`).

isabelroses:
> symbol picker of a [Nil](https://github.com/oxalica/nil)

This is a bit of a problem but nixd on the other hand should be able to handle this. Although nixd is not preferable if your using lix.

getchoo:
Finished in https://github.com/catppuccin/nix/pull/371

PR #371, opened 2024-11-10
refactor: move to catppuccin namespace

isabelroses:
- introduces a new lib function `lib.ctp.mkCatppuccinRenamedOpt`
- moves all options to `catppuccin.*`
- currently has a broken module in [zsh-syntax-highlighting.nix](https://github.com/catppuccin/nix/blob/821d694af86cbcfd7c99512117264310f6361876/modules/home-manager/zsh-syntax-highlighting.nix#L13)


modules/lib/default.nix
```diff
@@ -159,4 +159,36 @@ in
...
+          "flavor"
+        ]
+      )
+      [
+        "catppuccin"
+        to
+        "flavor"
+      ]
+    )
+  ];
```
getchoo, line 193: This is something I plan on adding to the actual mkCatppuccinOpt soon ™️, but it's probably good to start doing it here and get rid of some boilerplate in our modules

(It will also need to be reformatted, sorry I am lazy)

```suggestion
  mkCatppuccinRenamedOpt = {
    from,
    to,
    accentSupport ? false
  }:
  [
    (lib.mkRenamedOptionModule
      (
        from
        ++ [
          "enable"
        ]
      )
      [
        "catppuccin"
        to
        "enable"
      ]
    )

    (lib.mkRenamedOptionModule
      (
        from
        ++ [
          "flavor"
        ]
      )
      [
        "catppuccin"
        to
        "flavor"
      ]
    )
  ]
  ++ lib.optional accentSupport (
    lib.mkRenamedOptionModule
      (
        from
        ++ [
            "accent"
        ]
      )
      [
        "catppuccin"
        to
        "accent"
      ]
    );
```

modules/home-manager/zsh-syntax-highlighting.nix
```diff
@@ -2,14 +2,22 @@
...
     name = "Zsh Syntax Highlighting";
   };
 
+  # FIXME: this leads to an error
+  # imports = lib.ctp.mkCatppuccinRenamedOpt [
+  #   "programs"
+  #   "zsh"
+  #   "syntaxHighlighting"
+  #   "catppuccin"
+  # ] "zsh-syntax-highlighting";
```
getchoo, line 19: What is the error here? 🙀

getchoo (changes requested):
Most of this looks good

Only other change I suggest are changing the commit prefix to `feat` to trigger a minor release. We also need to check how the docs handles this renaming (we may need to change how we handle that too :p)

modules/home-manager/zsh-syntax-highlighting.nix
isabelroses: It results in this
> error: evaluation aborted with the following error message: 'cannot find attribute `programs.zsh.syntaxHighlighting.catppuccin.flavor''

modules/home-manager/kvantum.nix
```diff
@@ -43,6 +43,31 @@ in
...
+      to = "kvantum";
+      accentSupport = true;
+    })
+    ++ [
+      (lib.mkRenamedOptionModule
+        [
+          "qt"
+          "catppuccin"
+          "apply"
+        ]
```
arminius-smh, line 62: the option that has to be renamed is 'qt.style.catppuccin.apply'

Weathercold:
Closes #222

modules/home-manager/obs-studio.nix
```diff
@@ -1,13 +1,22 @@
 { config, lib, ... }:
 let
   inherit (config.catppuccin) sources;
-  cfg = config.programs.obs-studio.catppuccin;
+  cfg = config.catppuccin.obs;
```
Weathercold, line 4: Why change from `obs-studio` to `obs`? We should be consistent with home-manager's names.

modules/nixos/tty.nix
```diff
@@ -1,12 +1,20 @@
 { config, lib, ... }:
 let
   inherit (config.catppuccin) sources;
-  cfg = config.console.catppuccin;
+  cfg = config.catppuccin.tty;
```
Weathercold, line 4: again, I think `console` is better because it's the namespace NixOS uses. From a user's perspective, most don't know that the name of the catppuccin port is `tty`, but they may have changed a few `console.*` options, so it follows that they expect the catppuccin option to be named `console`.

Weathercold (changes requested):

modules/home-manager/obs-studio.nix
isabelroses: when moving the discussion was brought up as to whether we should use home-manager names or ctp based names and we concluded we would use ctp port names. See the original description

> port names are preferred to nix names

modules/nixos/tty.nix
isabelroses: same as https://github.com/catppuccin/nix/pull/371#discussion_r1855162178

modules/home-manager/obs-studio.nix
getchoo: I think sticking with port names is a better approach. It helps maintain consistency with our option names, avoids weird cases where a home-manager equivalent doesn't exist (see [`glamour`](https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/glamour.nix#L8)), and just makes more sense to me as a project in the org -- we should be thinking of things from a Catppuccin user's perspective rather than NixOS/home-manager

getchoo (changes requested):

modules/home-manager/cursor.nix
Weathercold, line 1: rename to `cursors.nix`

modules/home-manager/obs-studio.nix
Weathercold, line 1: rename to `obs.nix`

Weathercold (changes requested):

modules/home-manager/zsh-syntax-highlighting.nix
Weathercold: My guess is that `programs.zsh.syntaxHighlighting` is defined as a submodule. I tried to define renamed options in `programs.jq.colors` which is also a submodule and got the same error

isabelroses: Yeah seems so, https://github.com/nix-community/home-manager/blob/67cd4814a247fd0fe97171acb90659f7e304bcb8/modules/programs/zsh.nix#L238.

How would we get around this? Perhaps leaving the old options and including a warning?

Weathercold: Yes, try to manually define the renamed options and the warning and set the new options to the old ones by default. Basically replicating `mkRenamedOptionModule`

isabelroses: addressed https://github.com/catppuccin/nix/compare/4c7f5b65973d76448834398f221692cf54ca24b9..0c1924eb253c0c0eb15178e24b52d5fb285982fc

Weathercold: Did you try to set the new options to the old ones by default?

modules/home-manager/zsh-syntax-highlighting.nix
```diff
@@ -1,18 +1,39 @@
...
-    name = "Zsh Syntax Highlighting";
+  options = {
+    catppuccin.zsh-syntax-highlighting = ctp.mkCatppuccinOpt {
+      name = "Zsh Syntax Highlighting";
+      enableDefault = config.catppuccin.enable || oldCfg.enable;
+    };
+
+    programs.zsh.syntaxHighlighting.catppuccin = ctp.mkCatppuccinOpt {
+      name = "Zsh Syntax Highlighting";
+    };
```
Weathercold, line 23: set `enableDefault` to `false` so that the warning doesn't get triggered if `catppuccin.enable` is set to `true`

```suggestion
      enableDefault = false;
    };
```

tests/home.nix
```diff
@@ -34,9 +38,6 @@
       delta.enable = true;
     };
     gitui.enable = true;
-    # this is enabled by default already, but still
-    # listing explicitly so we know it's tested
-    glamour.catppuccin.enable = true;
```
getchoo, line 39: We should keep using this attribute in our tests to make sure the alias works (similar to the others)

modules/home-manager/neovim.nix
```diff
@@ -5,11 +5,20 @@
   ...
 }:
 let
-  cfg = config.programs.neovim.catppuccin;
+  cfg = config.catppuccin.neovim;
   enable = cfg.enable && config.programs.neovim.enable;
 in
 {
-  options.programs.neovim.catppuccin = lib.ctp.mkCatppuccinOpt { name = "neovim"; };
+  options.catppuccin.neovim = lib.ctp.mkCatppuccinOpt { name = "neovim"; };
```
getchoo, line 12: Should we rename this to `nvim`? That's what the port repo is called after all 🤔

modules/home-manager/zsh-syntax-highlighting.nix
```diff
@@ -1,18 +1,40 @@
...
+      initExtra = lib.mkBefore ''
+        source '${sources.zsh-syntax-highlighting}/themes/catppuccin_${cfg.flavor}-zsh-syntax-highlighting.zsh'
+      '';
+    };
+
+    catppuccin.zsh-syntax-highlighting.flavor = lib.mkDefault oldCfg.flavor;
+
+    warnings = lib.optional oldCfg.enable ''
+      `programs.zsh.syntaxHighlighting.catppuccin.enable` is deprecated. Please use `catppuccin.zsh-syntax-highlighting.enable` instead.
     '';
```
getchoo, line 38:
```suggestion
  config = lib.mkIf enable (lib.mkMerge [
    {
      programs.zsh = {
        initExtra = lib.mkBefore ''
          source '${sources.zsh-syntax-highlighting}/themes/catppuccin_${cfg.flavor}-zsh-syntax-highlighting.zsh'
        '';
      };

      catppuccin.zsh-syntax-highlighting.flavor = lib.mkDefault oldCfg.flavor;
    }
    
    (lib.mkIf oldCfg.enable {
      warnings = [ "programs.zsh.syntaxHighlighting.catppuccin.enable` is deprecated. Please use `catppuccin.zsh-syntax-highlighting.enable` instead." ];
    })
  ]);
```

getchoo (changes requested):
This should also be rebased to fix CI

getchoo (commented):
This needs another rebase after https://github.com/catppuccin/nix/pull/390 ~~hopefully it doesn't blow up our new search~~

isabelroses:
Now the zsh-syntax-highlighting will not show up too. Which is great.

image

modules/home-manager/zsh-syntax-highlighting.nix
getchoo, line 1: Looking into https://github.com/NixOS/nixpkgs/pull/360443 as a more proper way to handle this. We may need to pollyfill it until later in 24.11's life (assuming it's backported (or merged)), but I think it'd be nicer than the manual declaration here

modules/home-manager/zsh-syntax-highlighting.nix
```diff
@@ -1,18 +1,51 @@
...
+      lib.recursiveUpdate
+        (ctp.mkCatppuccinOpt {
+          name = "Zsh Syntax Highlighting";
+          enableDefault = false;
+        })
+        # hide the options from the search
+        {
+          enable.internal = true;
+          flavor.internal = true;
+        };
```
getchoo, line 31:
```suggestion
    programs.zsh.syntaxHighlighting = lib.mkOption {
      type = lib.types.submodule {
        options.catppuccin =
          ctp.mkCatppuccinOpt {
            name = "Zsh Syntax Highlighting";
            enableDefault = false;
          }
          // {
            # hide the options from the search
            enable.internal = true;
            flavor.internal = true;
          };
      };
    };
```
We should also be able to extend the submodule like so. I don't *think* it makes much difference here, but just in case anything breaks

getchoo (approved):
Thanks for all of your work here :)

khaneliman:
Rename options don't appear to be set up properly 
```bash
 error: The option `home-manager.users.khaneliman.wayland.windowManager.hyprland.catppuccin.accent' does not exist. Definition values:
       - In `/nix/store/18zdkdp786dwgl3ivji0a223n2gz13a0-iv22jisi34qdx0frafdgrlncz09nnr2r-source/modules/home/theme/catppuccin/default.nix':

PR #377, opened 2024-11-14
feat(home-manager): cava,imv,mako,swaylock sans IFD

### LINKED ISSUE #340: swaylock sans IFD
Sorry, I'm not going to look into this myself. Just want to get notified if/when :sweat_smile:.

### LINKED ISSUE #341: mako sans IFD
Sorry, I'm not going to look into this myself. Just want to get notified if/when 😅.

### LINKED ISSUE #373: fromINI sans IFD
If we're not aiming for actual INI parsing (is there even a formal spec?), parsing [these](https://github.com/catppuccin/swaylock/blob/77246bbbbf8926bdb8962cffab6616bc2b9e8a06/themes/macchiato.conf) in Nix shouldn't take too long.

Would such a change be accepted?

https://github.com/search?q=repo%3Acatppuccin%2Fnix%20fromini&type=code

- https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/swaylock.nix#L37
  https://github.com/catppuccin/swaylock/blob/77246bbbbf8926bdb8962cffab6616bc2b9e8a06/themes/macchiato.conf
- https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/imv.nix#L11
  https://github.com/catppuccin/imv/blob/0317a097b6ec8122b1da6d02f61d0c5158019f6e/themes/macchiato.config
- https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/mako.nix#L11-L13
  https://github.com/catppuccin/mako/blob/92844f144e72f2dc8727879ec141ffdacf3ff6a1/themes/catppuccin-frappe/catppuccin-frappe-blue
- https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/modules/home-manager/cava.nix#L14
  https://github.com/catppuccin/cava/blob/0746f77974330338ee2e1e4d1ef9872eba57eb26/themes/frappe.cava

Can these possibly be parsed using `fromTOML`?

### LINKED ISSUE #375: imv sans IFD
I'm working on this.

### LINKED ISSUE #376: cava sans IFD
I'm working on this.

mightyiam:
Closes #340
Closes #341
Closes #375
Closes #376
Closes #373

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,14 +182,7 @@ For [standalone installations](https://nix-community.github.io/home-manager/inde
   ```nix
   {
     programs = {
-      cava.catppuccin.enable = false;
       gh-dash.catppuccin.enable = false;
-      imv.catppuccin.enable = false;
-      swaylock.catppuccin.enable = false;
-    };
-  
-    services = {
-      mako.catppuccin.enable = false;
     };
   }
   ```
--- a/docs/src/faq.md
+++ b/docs/src/faq.md
@@ -20,14 +20,7 @@
   ```nix
   {
     programs = {
-      cava.catppuccin.enable = false;
       gh-dash.catppuccin.enable = false;
-      imv.catppuccin.enable = false;
-      swaylock.catppuccin.enable = false;
-    };
-  
-    services = {
-      mako.catppuccin.enable = false;
     };
   }
   ```
--- a/modules/home-manager/cava.nix
+++ b/modules/home-manager/cava.nix
@@ -11,6 +11,6 @@ in
   };
 
   config.programs.cava = lib.mkIf enable {
-    settings = lib.ctp.fromINIRaw (sources.cava + "/themes/${flavor}.cava");
+    settings = lib.ctp.fromINI (builtins.readFile (sources.cava + "/themes/${flavor}.cava"));
   };
 }
--- a/modules/home-manager/imv.nix
+++ b/modules/home-manager/imv.nix
@@ -8,6 +8,6 @@ in
   options.programs.imv.catppuccin = lib.ctp.mkCatppuccinOpt { name = "imv"; };
 
   config.programs.imv.settings = lib.mkIf enable (
-    lib.ctp.fromINI (sources.imv + "/themes/${cfg.flavor}.config")
+    lib.ctp.fromINI (builtins.readFile (sources.imv + "/themes/${cfg.flavor}.config"))
   );
 }
--- a/modules/home-manager/mako.nix
+++ b/modules/home-manager/mako.nix
@@ -8,9 +8,9 @@ let
   inherit (config.catppuccin) sources;
   cfg = config.services.mako.catppuccin;
   enable = cfg.enable && config.services.mako.enable;
-  theme = lib.ctp.fromINI (
+  theme = lib.ctp.fromINI (builtins.readFile (
     sources.mako + "/themes/catppuccin-${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}"
-  );
+  ));
 
   # Settings that need to be extracted and put in extraConfig
   extraConfigAttrs = lib.attrsets.getAttrs [ "urgency=high" ] theme;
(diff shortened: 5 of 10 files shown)


isabelroses (approved):
Thanks for the PR! 

My only real concern is maintainability. Seeing how its a bit more complex then anything we have currently and there is no spec, but rough commonalities. This will be especially apparent when we have more ports that are using INI. 

So I'm willing to approve but would like to ask @getchoo to confer.

mightyiam:
May I help with whatever is preventing this from merging?

getchoo (commented):
Sorry for the wait, this ended up a bit too far back in my inbox. Thanks for bumping it!

The implementation of `fromINI` here is really good and I would accept it. However, this isn't actually preventing IFD...or it shouldn't anyways

Currently we use [`npins`](https://github.com/andir/npins) to manage our sources. `npins` fetches these sources through the [builtin fetchers](https://github.com/catppuccin/nix/blob/32359bf226fe874d3b7a0a5753d291a4da9616fe/.sources/default.nix#L36), which unlike those found in Nixpkgs, block evaluation by downloading the paths -- just like a normal instance of IFD would. This isn't considered IFD by Nix itself though, as the builtin fetchers aren't technically outputting derivations.

To put it simply: our use of `npins` and builtin fetchers for our sources allows this

```nix
(builtins.readFile (sources.imv + "/themes/${cfg.flavor}.config")
```

to not be considered IFD, even though `builtins.readFile <expr>` is a [textbook example](https://nix.dev/manual/nix/2.25/language/import-from-derivation.html) of IFD. The solution here only takes advantage of this to avoid the requirement of the `--allow-import-from-derivation` flag, while still having the same blocking effects on evaluation; it will also no longer work once we move to Nixpkgs fetchers. This makes it very difficult to justify bringing this function in tree, as the only difference soon will be in this being a pure Nix solution rather than calling an external command -- which while cool in theory, doesn't bring much practical benefit here

I strongly encourage sharing this implementation though. For local files it would work wonders, and I'm sure prevent IFD in many other projects. `nixpkgs.lib` may be a good place :)

mightyiam:
> The solution here only takes advantage of this to avoid the requirement of the `--allow-import-from-derivation` flag

Sorry, I lost you there. I'm not sure what the meaning of "here" is, nor what "this" is.

getchoo:
"here" is referring to this pull request. "this" is referring to the way I described builtin fetchers blocking evaluation similar to IFD, without technically causing IFD

mightyiam:
So I am to understand that a change is being considered and possibly worked on, that will replace the use of builtin fetchers with nixpkgs fetchers. And that a result of that change is that IFD will be unavoidable for all modules?

getchoo:
> So I am to understand that a change is being considered and possibly worked on, that will replace the use of builtin fetchers with nixpkgs fetchers

Yes

> And that a result of that change is that IFD will be unavoidable for all modules?

No

Most modules that currently do not require IFD will continue to not -- in fact they will actually see a major improvement in the downloads of their sources no longer blocking evaluation. Some modules that take advantage of the use of builtin fetchers to *technically* not perform IFD will now require it, though. The changes in this PR are an example of the latter case.

mightyiam:
Okay, I understand enough that I would like to ask some more about it, but is there a better thread for that, please?

getchoo:
The introductory paragraph of the ["Fetchers"](https://nixos.org/manual/nixpkgs/stable/#chap-pkgs-fetchers) chapter in the Nixpkgs manual is a good resource. We've also recently opened [discussions](https://github.com/catppuccin/nix/discussions) for this repository, so feel free to open a thread there :)

mightyiam:
Thank you. May I see this WIP upcoming change, please?

getchoo:
https://github.com/catppuccin/nix/pull/384

PR #390, opened 2024-11-29
docs: use nuscht search for options

getchoo:
[NuschtOS/search](https://github.com/NuschtOS/search) is a simple and fast static-page NixOS option search. It implements a good amount of the logic we have been handling ourselves, while also giving a much better search interface that's specifically defined for module options. This is meant to replace our previous mdbook implementation (while still keeping other sections of our docs in mdbook)

It's currently a bit of a maintenance nightmare to be having this repository in a flake.lock of...this repository, but it'll be better after we remove the subflake in https://github.com/catppuccin/nix/pull/384 (and by then my PRs to remove extra inputs like flake-utils from NuschtOS/search will hopefully be through to lessen the blow). We also might want to look into Catppuccin-ifying the CSS introduced here. Other than that, pretty great upgrade :)


getchoo:
Can be tested with `nix run ./dev#serve`

isabelroses (approved):

PR #410, opened 2024-12-19
fix(home-manager/hyprland): rename accent

isabelroses:
re: https://github.com/catppuccin/nix/pull/371#issuecomment-2555603718

@khaneliman are you willing to test?


khaneliman:
Fixes error on my side, thanks for the quick response!

Issue #428, opened 2024-12-28
gtk: 'anonymous lambda' called with unexpected argument 'variant'


sofiedotcafe:
This piece of code fails to build:
```nix
gtk = {
  enable = true;
  catppuccin = {
    enable = true;
    icon.enable = true;
    gnomeShellTheme = true;
  };
};
```
... with the error as the following:
```
       … while evaluating the option `gtk.theme.package':

       … while evaluating definitions from `/nix/store/rgwcvkpzdhy6rqpn6509a1qwqjn6rg5f-source/modules/home-manager/gtk.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: function 'anonymous lambda' called with unexpected argument 'variant'
       at /nix/store/rgwcvkpzdhy6rqpn6509a1qwqjn6rg5f-source/pkgs/gtk/package.nix:1:1:
            1| {
             | ^
            2|   lib,
```

Has `catppuccin-nix` made any changes to ugh, `gtk` stuff recently? I took a quick look at the code and saw that this is probably an regression of https://github.com/catppuccin/nix/commit/115c3de5635c257bd2a723e06f8262a5edd66d9c due to the code using the wrong variable name for setting the package flavor: https://github.com/catppuccin/nix/blob/a2e641bc6b17129d81d54019e14c9956784c69c6/modules/home-manager/gtk.nix#L163 ... and which is defined in the package as: https://github.com/catppuccin/nix/blob/a2e641bc6b17129d81d54019e14c9956784c69c6/pkgs/gtk/package.nix#L9

PR #429, opened 2024-12-28
fix(home-manager/gtk): pass flavor and not variant

isabelroses:
closes https://github.com/catppuccin/nix/issues/428


isabelroses:
@sofiedotcafe do you mind confirm this works.

sofiedotcafe:
> @sofiedotcafe do you mind confirm this works.

Yes, it works!

PR #432, opened 2024-12-30
fix(pkgs/nvim): copy overrides from nixpkgs

### LINKED ISSUE #431: neovim plugin is broken with nixos-unstable override
to reproduce, simply enable neovim and have a flake with something similar to these inputs:
```nix
  inputs = {
    # points to 634fd46 at the time of writing this
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";

    catppuccin = {
      url = "github:catppuccin/nix";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        # disable inputs we don't need
        nixpkgs-stable.follows = "";
        home-manager.follows = "";
        home-manager-stable.follows = "";
        nuscht-search.follows = "";
        catppuccin-v1_1.follows = "";
        catppuccin-v1_2.follows = "";
      };
    };
```

i'm afraid i currently don't have the resources to do a full nixpkgs bisect since i can't build every derivation due to computing constraints, but [c207a2b](https://github.com/NixOS/nixpkgs/commit/c207a2becd4d89c5ffbadaed40e74bc6d64e128b) should be a bad commit while [d3c42f1](https://github.com/NixOS/nixpkgs/commit/d3c42f187194c26d9f0309a8ecc469d6c878ce33) (currently in the lockfile) is good, to narrow things down at least a bit.

[here](https://github.com/user-attachments/files/18271512/log.txt) is the full log output, but the following lines are probably sufficient:
```
All lua modules were checked.

Require check failed for the following modules:
  - catppuccin.groups.integrations.feline
  - catppuccin.groups.integrations.noice
  - catppuccin.lib.vim.init
```
nixpkgs seems to just disble these modules, as can be seen [here](https://github.com/NixOS/nixpkgs/blob/1a7de5d740a244b99c53e6bff8c60b621637f687/pkgs/applications/editors/vim/plugins/overrides.nix#L276-L282); i'll open a pr adding the override's changes as from my testing.

however, all of this feels rather weird (i have no clue why the plugin suddenly breaks), i barely know what any of this stuff does and there might be better solutions.

w-lfchen:
closes #431.

as pointed out in the issue, this works and i just copied [these lines](https://github.com/NixOS/nixpkgs/blob/1a7de5d740a244b99c53e6bff8c60b621637f687/pkgs/applications/editors/vim/plugins/overrides.nix#L276-L282), but i don't fully understand what the exact problem is or why this patch fixes it.
i don't know whether this is feasible, but it might be better to get these overrides from nixpkgs somehow. i can look into that, should it be deemed necessary.

### DIFF
--- a/pkgs/nvim/package.nix
+++ b/pkgs/nvim/package.nix
@@ -14,6 +14,12 @@ vimUtils.buildVimPlugin rec {
 
   src = sources.${portName};
 
+    nvimSkipModule = [
+      "catppuccin.groups.integrations.noice"
+      "catppuccin.groups.integrations.feline"
+      "catppuccin.lib.vim.init"
+    ];
+
   meta = {
     description = "Soothing pastel theme for ${portName}";
     homepage = "https://github.com/catppuccin/${portName}";


isabelroses (approved):
Thanks for the fix.

PR #433, opened 2024-12-30
feat(home-manager): add support for wezterm

taranarmo:
This adds support for WezTerm theme. There are no specific option fo themes in the wezterm home-manager module so I used `extraConfig`.

### DIFF
--- a/modules/home-manager/wezterm.nix
+++ b/modules/home-manager/wezterm.nix
@@ -4,18 +4,23 @@
 let
   inherit (config.catppuccin) sources;
   cfg = config.catppuccin.wezterm;
+  applyWeztermTheme = config.catppuccin.wezterm.apply or false;
 in
-
 {
   options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; };
 
+  options.catppuccin.wezterm.apply = lib.mkOption {
+    type = lib.types.bool;
+    default = false;
+    description = "Apply Catppuccin theme to WezTerm.";
+  };
 
-  config = lib.mkIf cfg.enable {
+  config = lib.mkIf (cfg.enable && applyWeztermTheme) {
     programs.wezterm = {
       colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
       extraConfig = lib.mkBefore ''
         local config = {}
-        if wezterm.c_builder then
+        if wezterm.config_builder then
           config = wezterm.config_builder()
         end
 
--- a/modules/home-manager/wezterm.nix
+++ b/modules/home-manager/wezterm.nix
@@ -4,28 +4,27 @@
 let
   inherit (config.catppuccin) sources;
   cfg = config.catppuccin.wezterm;
-  applyWeztermTheme = config.catppuccin.wezterm.apply or false;
 in
 {
-  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; };
-
-  options.catppuccin.wezterm.apply = lib.mkOption {
-    type = lib.types.bool;
-    default = false;
-    description = "Apply Catppuccin theme to WezTerm.";
+  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; } // {
+    apply = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      description = "Apply Catppuccin theme to WezTerm.";
+    };
   };
 
-  config = lib.mkIf (cfg.enable && applyWeztermTheme) {
+  config = lib.mkIf cfg.enable {
     programs.wezterm = {
       colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
-      extraConfig = lib.mkBefore ''
+      extraConfig = lib.mkIf cfg.apply (lib.mkBefore ''
         local config = {}
         if wezterm.config_builder then
           config = wezterm.config_builder()
         end
 
         dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
-      '';
+      '');
     };
   };
 }
--- a/modules/home-manager/wezterm.nix
+++ b/modules/home-manager/wezterm.nix
@@ -17,13 +17,16 @@ in
   config = lib.mkIf cfg.enable {
     programs.wezterm = {
       colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
-      extraConfig = lib.mkIf cfg.apply (lib.mkBefore ''
-        local config = {}
-        if wezterm.config_builder then
-          config = wezterm.config_builder()
-        end
+      extraConfig = ''
+        local catppuccin_plugin = "${sources.wezterm}/plugin/init.lua"
+      ''
+      + lib.mkIf cfg.apply (lib.mkBefore ''
+          local config = {}
+          if wezterm.config_builder then
+            config = wezterm.config_builder()
+          end
 
-        dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
+          dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
       '');
     };
   };
--- a/modules/home-manager/wezterm.nix
+++ b/modules/home-manager/wezterm.nix
@@ -20,14 +20,14 @@ in
       extraConfig = ''
         local catppuccin_plugin = "${sources.wezterm}/plugin/init.lua"
       ''
-      + lib.mkIf cfg.apply (lib.mkBefore ''
+      + lib.optionalString cfg.apply ''
           local config = {}
           if wezterm.config_builder then
             config = wezterm.config_builder()
           end
 
           dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
-      '');
+      '';
     };
   };
 }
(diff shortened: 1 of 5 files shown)


modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,29 @@
...
+    to = "wezterm";
+  };
+
+  config = lib.mkIf cfg.enable {
+    programs.wezterm = {
+      extraConfig = ''
+        return {
+          color_scheme = "Catppuccin ${catppuccinLib.mkUpper cfg.flavor}",
+        }
+      '';
```
isabelroses, line 26: although this is nice, this would break alot of configs beacuse we couldn't set other options. Instead we should use something like this, though that would enforce users to fit into their configuration.
```suggestion
      extraConfig = lib.mkBefore ''
        local config = {}
        if wezterm.c_builder then
          config = wezterm.config_builder()
        end
       
        require("${sources.wezterm}/init.lua").apply_to_config(config)
      '';
```

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,29 @@
...
+    from = [
+      "programs"
+      "wezterm"
+      "catppuccin"
+    ];
+    to = "wezterm";
+  };
+
+  config = lib.mkIf cfg.enable {
+    programs.wezterm = {
```
isabelroses, line 21:
```suggestion
    programs.wezterm = {
      colorSchemes = lib.importTOML "${sources.wezterm}/catppuccin-${cfg.flavor}.toml";
```

isabelroses (commented):
My suggestions do depend on a catppuccin wezterm package.

modules/home-manager/wezterm.nix
taranarmo: Yeah, I tried to add more config after PR creation and it all broke. I believe your solution is the best until home-manager module will be expanded with proper configuration options.

isabelroses (commented):
using `config.catppuccin.sources.wezterm` will require a package, take a look at https://github.com/catppuccin/nix/blob/main/CONTRIBUTING.md.

pkgs/wezterm/package.nix
```diff
@@ -0,0 +1,12 @@
...
+
+buildCatppuccinPort {
+  pname = "wezterm";
+
+  dontCatppuccinInstall = true;
+
+  postInstall = ''
+    mkdir -p $out
+    mv *.toml $out/
+  '';
```
isabelroses, line 11: I think this will be better, and should allow for the `require("${sources.wezterm}/init.lua").apply_to_config(config)` line to work :)
```suggestion
  installTargets = [ "dist" "plugin" ];
```

taranarmo: I did that (and updated paths to the `toml` files and `init.lua`) though now wezterm can't find the init.lua though it is accessible at the location (on nix darwin). No idea why.

isabelroses: ahh, sorry that would be partly my fault in getting used to the new package system. you will have to specify `dist/path-to.toml` and `plugin/init.lua` instead of what we have currently

isabelroses: but it seems you figured that out

isabelroses: ahh the user would also have to add `return config`, which i left out incase a user wanted to add more code.

taranarmo:
Eventually with help of search engines and chatbots I was able to make it working. I replaced `require` with `dofile` which solved the issue.

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,26 @@
...
+  cfg = config.catppuccin.wezterm;
+in
+
+{
+  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; };
+
+
+  config = lib.mkIf cfg.enable {
+    programs.wezterm = {
+      colorSchemes = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
```
isabelroses, line 15: I just noticed its a attrs of toml's
```suggestion
      colorSchemes."catppuccin-${cfg.flavour}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
```

isabelroses (commented):
overall, LGTM.

taranarmo:
Rebased and resolved conflicts

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,26 @@
...
+    programs.wezterm = {
+      colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
+      extraConfig = lib.mkBefore ''
+        local config = {}
+        if wezterm.c_builder then
+          config = wezterm.config_builder()
+        end
+
+        dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
+      '';
```
isabelroses, line 23: I was thinking about this, and how its a somewhat weird as an API for us to provide. Perhaps we should put this under a second option like `options.catppuccin.wezterm.apply`?

taranarmo: I added the option, please take a look

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,31 @@
...
+{
+  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; };
+
+  options.catppuccin.wezterm.apply = lib.mkOption {
+    type = lib.types.bool;
+    default = false;
+    description = "Apply Catppuccin theme to WezTerm.";
+  };
+
+  config = lib.mkIf (cfg.enable && applyWeztermTheme) {
```
isabelroses, line 18:
```suggestion
  config = lib.mkIf cfg.enable {
```

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,31 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.catppuccin.wezterm;
+  applyWeztermTheme = config.catppuccin.wezterm.apply or false;
```
isabelroses, line 7:
```suggestion
```

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,31 @@
...
+in
+{
+  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; };
+
+  options.catppuccin.wezterm.apply = lib.mkOption {
+    type = lib.types.bool;
+    default = false;
+    description = "Apply Catppuccin theme to WezTerm.";
+  };
```
isabelroses, line 16:
```suggestion
  options.catppuccin.wezterm = catppuccinLib.mkCatppuccinOption { name = "wezterm"; } // {
    apply = lib.mkOption {
      type = lib.types.bool;
      default = false;
      description = "Apply Catppuccin theme to WezTerm.";
    };
  };
```

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,31 @@
...
+    programs.wezterm = {
+      colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
+      extraConfig = lib.mkBefore ''
+        local config = {}
+        if wezterm.config_builder then
+          config = wezterm.config_builder()
+        end
+
+        dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
+      '';
```
isabelroses, line 28:
```suggestion
      extraConfig = lib.mkIf cfg.apply (lib.mkBefore ''
        local config = {}
        if wezterm.config_builder then
          config = wezterm.config_builder()
        end

        dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
      '');
```

isabelroses (changes requested):
Perhaps something more like this

isabelroses (approved):

taranarmo:
I updated the code though now it doesn't follow the global `catppuccin.enable = true`. Should we enable the `apply` by default?

isabelroses:
> I updated the code though now it doesn't follow the global `catppuccin.enable = true`. Should we enable the `apply` by default?

I think not. Because this will often be a breaking change and its a bit of a strange to use. we do something similar to this for kvantum.

taranarmo:
> > I updated the code though now it doesn't follow the global `catppuccin.enable = true`. Should we enable the `apply` by default?
> 
> I think not. Because this will often be a breaking change and its a bit of a strange to use. we do something similar to this for kvantum.

Okay then, we can update the code once the wezterm module for home manager will be updated to support more than `extraConfig`. Thanks for your comments and suggestions!

getchoo (approved):
Diff LGTM

modules/home-manager/wezterm.nix
```diff
@@ -0,0 +1,30 @@
...
+  config = lib.mkIf cfg.enable {
+    programs.wezterm = {
+      colorSchemes."catppuccin-${cfg.flavor}" = lib.importTOML "${sources.wezterm}/dist/catppuccin-${cfg.flavor}.toml";
+      extraConfig = lib.mkIf cfg.apply (lib.mkBefore ''
+        local config = {}
+        if wezterm.config_builder then
+          config = wezterm.config_builder()
+        end
+
+        dofile("${sources.wezterm}/plugin/init.lua").apply_to_config(config)
```
getchoo, line 26: Might want to export this as a variable so it's actually able to be imported in someone's own config, though

Maybe something like
```nix
extraConfig = ''
  local catppuccin_plugin = "${sources.wezterm}/plugin/init.lua"
''
+ lib.mkIf cfg.apply (lib.mkBefore ''
  <...?
'');
```

taranarmo: Updated, please take a look. Lua isn't my strong side as well as Nix 😅

getchoo (approved):

Issue #420, opened 2024-12-21
Support separate light/dark themes


samhh:
I've migrated my `programs.zed-editor` to take advantage of #359 (thank you!) and had to override the light theme given currently catppuccin only supports a single "flavour" which is implicitly applied across both light and dark modes.

Ideally, given the likes of Zed support specifying separate light and dark themes, we'd be able to optionally configure a different flavour for each. Presumably then there'd need to be some notion of a default where that dichotomy doesn't exist.

isabelroses:
It would be cool to see this. Though it would be interesting. One thing to note is only some ports would be capable of using the dark/light themes. Which may make it hard to implement.

A potential idea would be to modify the `catppuccinLib.mkCatppuccinOptions` to handle this, however it would make a bit of a grey area around the global `flavour` option.

Prior art in userstyles:
https://github.com/catppuccin/userstyles/blob/f90308af39acd7843372d1134965a826d4aa5534/template/catppuccin.user.css#L13-14

42willow:
in my opinion ports that support both light and dark themes should include latte as the light mode variant by default, but there should also be a global or local option to disable this (resulting in the global `flavour` option being used for both light and dark)

for example:
`catppuccin.lightDarkEnable = true;`
```nix
theme = {
  light = "Catppuccin Latte";
  dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
};
```

`catppuccin.lightDarkEnable = false;`
```nix
theme = {
  dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
  dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
};
```

this could of course be shortened with an if statement but I don't know nix

uncenter:
> in my opinion ports that support both light and dark themes should include latte as the light mode variant by default, but there should also be a global or local option to disable this (resulting in the global `flavour` option being used for both light and dark)
> 
> for example: `catppuccin.lightDarkEnable = true;`
> 
> ```nix
> theme = {
>   light = "Catppuccin Latte";
>   dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
> };
> ```
> 
> `catppuccin.lightDarkEnable = false;`
> 
> ```nix
> theme = {
>   dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
>   dark = "Catppuccin " + catppuccinLib.mkUpper cfg.flavor;
> };
> ```
[...]

PR #435, opened 2024-12-30
feat(home-manager/zed): accent support

isabelroses:


modules/home-manager/zed-editor.nix
```diff
@@ -2,30 +2,47 @@
...
-  };
 
   config = lib.mkIf cfg.enable {
     programs.zed-editor = {
-      extensions = [ "catppuccin" ];
-
       userSettings.theme = {
         light =
           "Catppuccin "
           + catppuccinLib.mkUpper cfg.flavor
```
42willow, line 27:
```suggestion
          "Catppuccin Latte"
```

modules/home-manager/zed-editor.nix
```diff
@@ -2,30 +2,47 @@
...
+          + "("
```
42willow, line 28:
```suggestion
          + " ("
```

modules/home-manager/zed-editor.nix
42willow: `programs.zed-editor.userSettings.theme.mode = dark` can be used to override to dark

42willow (changes requested):
thanks!

modules/home-manager/zed-editor.nix
42willow: actually never mind: see #420

modules/home-manager/zed-editor.nix
```diff
@@ -2,30 +2,47 @@
...
+          + " ("
+          + cfg.accent
+          + ")"
           + lib.optionalString (!cfg.italics) " - No Italics";
       };
     };
+
+    xdg.configFile = {
+      "zed/themes/catppuccin.json".source = "${sources.zed}/catppuccin-${
+        lib.optionalString (!cfg.italics) "-no-italics-"
```
42willow, line 44:
```suggestion
        lib.optionalString (!cfg.italics) "no-italics-"
```

42willow (changes requested):

42willow (approved):
all working! tysm :)

Issue #434, opened 2024-12-30
fcitx5 themes link broken


IsaacWangTT:
https://github.com/catppuccin/nix/blob/c56711a2636881ffcf70d2c23198eb98523d8016/modules/home-manager/fcitx5.nix#L59

It should be modified to

```nix
source = "${sources.fcitx5}/share/fcitx5/themes/catppuccin-${cfg.flavor}-${cfg.accent}"; 
```

PR #436, opened 2024-12-30
fix(home-manager/fcitx5): ajust source

isabelroses:
closes https://github.com/catppuccin/nix/issues/434


isabelroses:
@IsaacWangTT do you mind confirming this works?

Issue #438, opened 2024-12-30
GTK module doesn't seem to work anymore


AlexNabokikh:
Hi Everyone,

After updating flake to the latest commit `c56711a2636881ffcf70d2c23198eb98523d8016` the GTK module doesn't work anymore. `assets`  `gtk-dark.css` and  `gtk.css` in `.config/gtk-4.0` are empty. 

Here is my home-manager module config:

```Nix
catppuccin = {
  flavor = "macchiato";
  accent = "lavender";
  gtk = {
    enable = true;
    gnomeShellTheme = true;
  };
};
```

<details> <summary>flake lock before upgrade (GTK works properly)</summary>

```sh
{
  "nodes": {
    "brew-src": {
      "flake": false,
      "locked": {
        "lastModified": 1731323744,
        "narHash": "sha256-SxUQm4cTHcaoPQHoXe26ZV8cZiMWBGow8MjE4L+MckM=",
        "owner": "Homebrew",
        "repo": "brew",
        "rev": "254bf3fe9d8fa2e1b2fb55dbcf535b2d870180c4",
        "type": "github"
      },
      "original": {
        "owner": "Homebrew",
        "ref": "4.4.5",
        "repo": "brew",
        "type": "github"
      }
    },
    "catppuccin": {
      "locked": {
        "lastModified": 1734671418,
        "narHash": "sha256-K2Su5hM1nEIgKJ55TB5W+UfN9zBHUxC0SjWAtjXLEnI=",
        "owner": "catppuccin",
        "repo": "nix",
        "rev": "6239449c96569547af621899f44a5ea333cb2576",
        "type": "github"
      },
      "original": {
        "owner": "catppuccin",
        "repo": "nix",
        "type": "github"
      }
    },
    "darwin": {
      "inputs": {
        "nixpkgs": [
          "nixpkgs"
        ]
      },
      "locked": {
        "lastModified": 1733570843,
        "narHash": "sha256-sQJAxY1TYWD1UyibN/FnN97paTFuwBw3Vp3DNCyKsMk=",
        "owner": "LnL7",
        "repo": "nix-darwin",
        "rev": "a35b08d09efda83625bef267eb24347b446c80b8",
        "type": "github"
      },
      "original": {
        "owner": "LnL7",
        "repo": "nix-darwin",
        "type": "github"
      }
    },
    "flake-compat": {
      "flake": false,
      "locked": {
        "lastModified": 1733328505,
[...]

PR #439, opened 2024-12-30
fix(home-manager/gtk): ajust config.gtk.name for new package

isabelroses:
closes https://github.com/catppuccin/nix/issues/438

@AlexNabokikh do you mind testing :)


AlexNabokikh:
That seems to work properly. Thank you very much for a such a prompt fix! 
A true Catppuccin MVP!

PR #442, opened 2024-12-31
feat: dark and light theme support

### LINKED ISSUE #420: Support separate light/dark themes

isabelroses:
Perhaps a oversight but a understandable one seeing as we currently only have 1 reactive theme, zed. But that may change with the addition of ghostty!

closes https://github.com/catppuccin/nix/issues/420

The idea here was to introduce `darkFlavor` and `lightFlavor` in a very none breaking way, such that they both default to `config.catppuccin.flavor` rather then `mocha` and `latte` respectively.

### DIFF
--- a/modules/global.nix
+++ b/modules/global.nix
@@ -25,6 +25,20 @@ in
       description = "Global Catppuccin flavor";
     };
 
+    darkFlavor = lib.mkOption {
+      type = catppuccinLib.types.flavor;
+      default = config.catppuccin.flavor;
+      defaultText = "catppuccin.flavor";
+      description = "Global Catppuccin dark flavor";
+    };
+
+    lightFlavor = lib.mkOption {
+      type = catppuccinLib.types.flavor;
+      default = config.catppuccin.flavor;
+      defaultText = "catppuccin.flavor";
+      description = "Global Catppuccin light flavor";
+    };
+
     accent = lib.mkOption {
       type = catppuccinLib.types.accent;
       default = "mauve";
--- a/modules/home-manager/zed-editor.nix
+++ b/modules/home-manager/zed-editor.nix
@@ -8,7 +8,6 @@ let
   enable = cfg.enable && config.programs.zed-editor.enable;
 
   accent = if cfg.accent == "mauve" then "" else " (${cfg.accent})";
-  flavor = catppuccinLib.mkFlavorName cfg.flavor;
 in
 
 {
@@ -16,6 +15,7 @@ in
     catppuccinLib.mkCatppuccinOption {
       name = "zed";
       accentSupport = true;
+      darkLightSupport = true;
     }
     // {
       italics = lib.mkEnableOption "the italicized version of theme" // {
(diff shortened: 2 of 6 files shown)


modules/global.nix
```diff
@@ -26,6 +26,18 @@ in
...
+      type = catppuccinLib.types.flavor;
+      default = config.catppuccin.flavor;
+      description = "Global Catppuccin dark flavor";
+    };
+
+    lightFlavor = lib.mkOption {
+      type = catppuccinLib.types.flavor;
+      default = config.catppuccin.flavor;
+      description = "Global Catppuccin light flavor";
+    };
```
getchoo, line 39: I'm not too big a fan of this API. I think it'd be better to have a single enum option defining whether or not to use dark mode, and continuing using the local `flavor` option in modules (which should be fine since both can still be overridden on a per-module basis if you'd like light mode with a different-than-global flavor for example)

i.e., this would become

```nix
{
  # Or maybe name it `mode`? I dunno
  theme = lib.mkOption {
    type = lib.types.oneOf [ "light" "dark" ];
    default = dark;
    description = "Global Catppuccin theme type";
    example = lib.literalExpression "light";
  };
}

File: modules/home-manager/zed-editor.nix [hunk 2]
```diff
@@ -19,11 +24,11 @@ in
       userSettings.theme = {
         light =
           "Catppuccin "
-          + catppuccinLib.mkUpper cfg.flavor
+          + catppuccinLib.mkUpper cfg.lightFlavor
```
getchoo, line 27: This would stay the same
```suggestion
          + catppuccinLib.mkUpper cfg.flavor
```

modules/home-manager/zed-editor.nix
```diff
@@ -19,11 +24,11 @@ in
...
           + lib.optionalString (!cfg.italics) " - No Italics";
         dark =
           "Catppuccin "
-          + catppuccinLib.mkUpper cfg.flavor
+          + catppuccinLib.mkUpper cfg.darkFlavor
```
getchoo, line 31: Ditto

modules/home-manager/zed-editor.nix
getchoo, line 1: And then we would configure zed to use the light or dark theme

...assuming that's possible. I'd really like it to be since I feel like making the light/dark theme choice imperative is a bit odd and unlike what we do with flavors, accents, and enabling the themes themselves

getchoo (changes requested):

modules/global.nix
isabelroses: I see the idea, but i assume some people like macOS users would be more wanting to have the dark and light options since they are not enforced into declarative light/dark.

isabelroses: Oh and totally forgot for ghostty which is a single string, making it a bit hard to do that.

getchoo: I'm not too sure how else to generalize this API then. Having dynamic light/dark is virtually exclusive to macOS, and most apps will just not support it; I don't know where else this would really be used besides Ghostty and Zed

If we can configure either of those two to use a "dynamic" mode, I would be fine with that being a third option in this enum -- and those two ports having their own exclusive light/darkFlavor options. Otherwise I don't think this is a good fit here

different-name:
> Perhaps a oversight but a understandable one seeing as we currently only have 1 reactive theme, zed. But that may change with the addition of ghostty!

If Firefox Userstyles gets merged, that'd be another - currently setting both light and dark to the same flavor

hackardoX:
I'm super excited about this feature! Any idea on how we're going to roll it out?

HeitorAugustoLN:
This would be really neat for #549, since COSMIC has support for automatically switching themes when it's day or night

avycado13:
any updates on this?

modules/global.nix
isabelroses: > Having dynamic light/dark is virtually exclusive to macOS

you can now have a light/dark switching on linux more modernly. so maybe we could achieve this now. and is becoming more common of a option like in fish. we also support nixos modules which are websites so they would naturally have light/dark toggles.

i think the enum isn't a great approach tho since it makes the api much more complex than it needs to be. and we would have to maybe introduce new options. for example the fish port supports dynamic theming but only with latte light mode. which compared to userstyles feels regressive.

Issue #444, opened 2024-12-31
document cachix


isabelroses:
we have a cachix but we haven't documented it leading to alot of people not using it

https://app.cachix.org/cache/catppuccin

PR #445, opened 2025-01-01
feat: global cachix option

isabelroses:
closes https://github.com/catppuccin/nix/issues/444

### DIFF
--- a/modules/global.nix
+++ b/modules/global.nix
@@ -46,7 +46,7 @@ in
         description = "Port sources used across all options";
       };
 
-    cache.enable = lib.mkEnableOption "Enable Catppuccin cache";
+    cache.enable = lib.mkEnableOption "the usage of Catppuccin's binary cache";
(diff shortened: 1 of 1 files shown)


modules/global.nix
```diff
@@ -45,5 +45,14 @@ in
         apply = lib.recursiveUpdate defaultSources;
         description = "Port sources used across all options";
       };
+
+    cache.enable = lib.mkEnableOption "Enable Catppuccin cache";
```
getchoo, line 49:
```suggestion
    cache.enable = lib.mkEnableOption "the usage of Catppuccin's binary cache";
```

getchoo (changes requested):
It'd also be nice to document this somewhere

but we also don't really document `catppuccin.enable` in the getting started guide so...maybe that can wait :p

getchoo (approved):

PR #446, opened 2025-01-03
feat(home-manager): add support for ghostty

### LINKED ISSUE #427: Add support for Ghostty
https://github.com/catppuccin/ghostty

natecox:
Fixes #427

### DIFF
--- a/modules/home-manager/ghostty.nix
+++ b/modules/home-manager/ghostty.nix
@@ -1,0 +1,25 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+
+  cfg = config.catppuccin.ghostty;
+  themeName = "catppuccin-${cfg.flavor}";
+  enable = cfg.enable && config.programs.ghostty.enable;
+in
+{
+  options.catppuccin.ghostty = catppuccinLib.mkCatppuccinOption { name = "ghostty"; };
+
+  config = lib.mkIf enable {
+    xdg.configFile = {
+      "ghostty/themes/${themeName}".source = "${sources.ghostty}/${themeName}.conf";
+    };
+
+    programs.ghostty = {
+      settings = {
+        theme = "light:${themeName},dark:${themeName}";
+      };
+    };
+  };
+}
--- a/pkgs/sources.json
+++ b/pkgs/sources.json
@@ -55,6 +55,10 @@
     "hash": "sha256-fOCZxrEyWLi+VYnx3QYOP1R+VBhllhOlnO5/5Wg5aq4=",
     "rev": "acb1b1c22446e34781731ddbfb5e9bd699eccc74"
   },
+  "ghostty": {
+    "hash": "sha256-YHHOGeXNlVdUdUXHQW4G+6Nl8xyHo2tuJ9b+/1LfGos=",
+    "rev": "1cb0c33770a411160cbe14b48b831455a2244ee3"
+  },
   "gitui": {
     "hash": "sha256-CRxpEDShQcCEYtSXwLV5zFB8u0HVcudNcMruPyrnSEk=",
     "rev": "c7661f043cb6773a1fc96c336738c6399de3e617"
--- a/flake.lock
+++ b/flake.lock
@@ -53,11 +53,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1736204492,
-        "narHash": "sha256-CoBPRgkUex9Iz6qGSzi/BFVUQjndB0PmME2B6eEyeCs=",
+        "lastModified": 1736508663,
+        "narHash": "sha256-ZOaGwa+WnB7Zn3YXimqjmIugAnHePdXCmNu+AHkq808=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "20665c6efa83d71020c8730f26706258ba5c6b2a",
+        "rev": "2532b500c3ed2b8940e831039dcec5a5ea093afc",
         "type": "github"
       },
       "original": {
@@ -73,11 +73,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1735344290,
-        "narHash": "sha256-oJDtWPH1oJT34RJK1FSWjwX4qcGOBRkcNQPD0EbSfNM=",
+        "lastModified": 1736373539,
+        "narHash": "sha256-dinzAqCjenWDxuy+MqUQq0I4zUSfaCvN9rzuCmgMZJY=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "613691f285dad87694c2ba1c9e6298d04736292d",
+        "rev": "bd65bc3cde04c16755955630b344bc9e35272c56",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -16,12 +16,12 @@
     },
     "catppuccin-v1_2": {
       "locked": {
-        "lastModified": 1734728407,
-        "narHash": "sha256-Let3uJo4YDyfqbqaw66dpZxhJB2TrDyZWSFd5rpPLJA=",
-        "rev": "23ee86dbf4ed347878115a78971d43025362fab1",
-        "revCount": 341,
+        "lastModified": 1734734291,
+        "narHash": "sha256-CFX4diEQHKvZYjnhf7TLg20m3ge1O4vqgplsk/Kuaek=",
+        "rev": "1e4c3803b8da874ff75224ec8512cb173036bbd8",
+        "revCount": 344,
         "type": "tarball",
-        "url": "https://api.flakehub.com/f/pinned/catppuccin/nix/1.2.0/0193e5e0-33b7-7149-a362-bfe56b20f64e/source.tar.gz"
+        "url": "https://api.flakehub.com/f/pinned/catppuccin/nix/1.2.1/0193e646-1107-7f69-a402-f2a3988ecf1d/source.tar.gz"
       },
       "original": {
         "type": "tarball",
@@ -53,11 +53,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1734622215,
-        "narHash": "sha256-OOfI0XhSJGHblfdNDhfnn8QnZxng63rWk9eeJ2tCbiI=",
+        "lastModified": 1736204492,
+        "narHash": "sha256-CoBPRgkUex9Iz6qGSzi/BFVUQjndB0PmME2B6eEyeCs=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "1395379a7a36e40f2a76e7b9936cc52950baa1be",
+        "rev": "20665c6efa83d71020c8730f26706258ba5c6b2a",
         "type": "github"
       },
       "original": {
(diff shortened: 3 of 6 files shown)


modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,18 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  cfg = config.catppuccin.ghostty;
+  enable = cfg.enable && config.programs.ghostty.enable;
```
isabelroses, line 6: This module is correctly gatekept upstream via the `config.programs.ghostty.enable`, so we don't need to copy it here.
```suggestion
```

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,18 @@
...
+in
+{
+  options.catppuccin.ghostty = catppuccinLib.mkCatppuccinOption { name = "ghostty"; };
+
+  config = lib.mkIf enable {
```
isabelroses, line 11:
```suggestion
  config = lib.mkIf cfg.enable {
```

pkgs/sources.json
getchoo, line 1: We should be using the theme file from this package by linking it in $XDG_CONFIG_DIR/ghostty/themes

natecox: I'm not sure I follow you. There was a request here to use the light:,dark: syntax https://github.com/catppuccin/nix/issues/427#issuecomment-2564360720

modules/home-manager/ghostty.nix
natecox: Done

modules/home-manager/ghostty.nix
natecox: Done

pkgs/sources.json
isabelroses: yes we should use that but we can also use `programs.ghostty.themes."catppuccin-${cfg.flavor}" = catppuccinLib.importINI "${config.catppuccin.sources.ghostty}/catppuccin-${cfg.flavor}.conf";`.

natecox: oh yeah I see what you mean there; I was under the impression that they were mutually exclusive for some reason

natecox: ok confirmed that the theme is now where it's supposed to be

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,22 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  cfg = config.catppuccin.ghostty;
```
getchoo, line 5:
```suggestion
  inherit (config.catppuccin) sources;

  cfg = config.catppuccin.ghostty;
  enable = cfg.enable && programs.ghostty.enable;
  
  themeName = "catppuccin-${cfg.flavor}";
  themeConfigFile = themeName + ".conf";
```

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,22 @@
...
+{
+  options.catppuccin.ghostty = catppuccinLib.mkCatppuccinOption { name = "ghostty"; };
+
+  config = lib.mkIf cfg.enable {
+    programs.ghostty = {
+      themes = {
+        "catppuccin-${cfg.flavor}" =
+          catppuccinLib.importINI "${config.catppuccin.sources.ghostty}/catppuccin-${cfg.flavor}.conf";
+      };
+
```
getchoo, line 16:
```suggestion
```

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,22 @@
...
+      settings = {
+        theme = "light:catppuccin-${cfg.flavor},dark:catppuccin-${cfg.flavor}";
+      };
+    };
```
getchoo, line 20:
```suggestion
    };
    
    xdg.configFile = {
      "ghostty/themes/${themeFile}".source = "${sources.ghostty}/${themeFile}";
    };
```

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,22 @@
...
+
+  config = lib.mkIf cfg.enable {
+    programs.ghostty = {
+      themes = {
+        "catppuccin-${cfg.flavor}" =
+          catppuccinLib.importINI "${config.catppuccin.sources.ghostty}/catppuccin-${cfg.flavor}.conf";
+      };
+
+      settings = {
+        theme = "light:catppuccin-${cfg.flavor},dark:catppuccin-${cfg.flavor}";
```
getchoo, line 18:
```suggestion
        theme = "light:${themeName},dark:${themeName}";
```

modules/home-manager/ghostty.nix
uncenter: I assume you mean `themeConfigFile` not `themeFile` here, but also I don't think the theme file under ghostty/themes is supposed to have a file extension, unlike the theme file from the source does.

getchoo: Ah yeah it'd just be `ghostty/themes/${themeName}` then

We could also use the absolute path apparently

> If the theme is an absolute pathname, Ghostty will attempt to load that file as a theme

and I *almost* suggested that, but it'd probably be better to link it here as to avoid rebuilding this configuration file every time the theme source updates lol

zakuciael:
Hi, is there any reason this PR was closed? Or if the author is no longer interested in implementing it, is it okay to make a new PR adding this module?

natecox:
> Hi, is there any reason this PR was closed? Or if the author is no longer interested in implementing it, is it okay to make a new PR adding this module?

Seemed like I was getting the runaround a bit on this without any real conversation happening to tell me why and I lost interest. 

I’d be happy to pick this back up if a clear direction can be provided, I’m just not sure what to do with two reviewers asking for conflicting changes. 

No hard feelings or anything, because text tends to present negatively.

getchoo:
We're sorry about that, genuinely

There have been a lot of large refactors recently, and its made us change the approach we use in applying ports to different programs. I'll admit I'm the one who pushed a lot of these changes, and I'm more familiar with the effects they have than @isabelroses. I should have communicated these better, as well as some of the new things to look out for; that's my bad

The issue that caused this difference in review -- [Import from Derivation](https://nix.dev/manual/nix/2.25/language/import-from-derivation.html) -- is one that has been a pretty hot topic recently, and it's no doubt because of this. I plan on writing an explainer in our docs so others aren't blind sided by this again, as it's not exactly easy to catch

I'd love to see this continued, and I'll make sure to answer any questions along the way

natecox:
All right, well I appreciate the context. Can you point out a port that is up to date that I can reference? I started with copying the kitty config (thinking it would make the port simple lol).

natecox:
@getchoo I think I've resolved all the requests above.

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,24 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+
+  cfg = config.catppuccin.ghostty;
```
getchoo, line 7:
```suggestion
  cfg = config.catppuccin.ghostty;
  enable = cfg.enable && programs.ghostty.enable;
```
Since we're also placing a file in `~/.config` with `xdg.configFile`, we should make sure that is only done when Ghostty itself is enabled (just like the values of `programs.ghostty.settings` only being written to its config file when its enabled)

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,24 @@
...
+  themeName = "catppuccin-${cfg.flavor}";
+in
+{
+  options.catppuccin.ghostty = catppuccinLib.mkCatppuccinOption { name = "ghostty"; };
+
+  config = lib.mkIf cfg.enable {
```
getchoo, line 13:
```suggestion
  config = lib.mkIf enable {
```
And the way we usually do that is just by guarding all of our changes behind that condition

getchoo (changes requested):
Otherwise this LGTM. I'll make sure to test it out soon since I know this is a popular app and port ;)

getchoo:
> Can you point out a port that is up to date that I can reference? 

Honestly, I think this is already a great example of how we want these modules to be handled now 😆 

But for the future, some of my favorites are

- [Alacritty](https://github.com/catppuccin/nix/blob/2dafc413dd280c2616524cd2d286bbb7c3e79cd5/modules/home-manager/alacritty.nix)
- [bat](https://github.com/catppuccin/nix/blob/2dafc413dd280c2616524cd2d286bbb7c3e79cd5/modules/home-manager/bat.nix)
- [btop](https://github.com/catppuccin/nix/blob/2dafc413dd280c2616524cd2d286bbb7c3e79cd5/modules/home-manager/btop.nix)
- [Hyprland](https://github.com/catppuccin/nix/blob/2dafc413dd280c2616524cd2d286bbb7c3e79cd5/modules/home-manager/hyprland.nix)
- [k9s](https://github.com/catppuccin/nix/blob/2dafc413dd280c2616524cd2d286bbb7c3e79cd5/modules/home-manager/k9s.nix)

With k9s and btop being the most similar to this one

The general idea is to place our port sources at the right path (or list exactly where they are) so that the program can read that at runtime. This is opposed to us reading these files at evaluation time in the modules to set options like `programs.someThemeableApp.settings.hexColor = "#FFFFFF";` -- as even though Nix totally can do that, it causes some small but nasty issues

modules/home-manager/ghostty.nix
natecox: I want to confirm this one, because I actually had that implemented as you suggest and was told it was unnecessary here: https://github.com/catppuccin/nix/pull/446#discussion_r1901586957

getchoo: At that revision the comment was correct due to the module only touching `programs.ghostty` options

Those are already only applied when `programs.ghostty.enable` is `true`. When we touch options outside of it (in this case `xdg.configFile`) though, we need to implement that same guard internally (else we may end up placing port files on people's systems for programs they're not actually using)

natecox: okie dokie.

modules/home-manager/ghostty.nix
```diff
@@ -0,0 +1,25 @@
...
+  config = lib.mkIf enable {
+    xdg.configFile = {
+      "ghostty/themes/${themeName}".source = "${sources.ghostty}/${themeName}.conf";
+    };
+
+    programs.ghostty = {
+      settings = {
+        theme = "light:${themeName},dark:${themeName}";
+      };
+    };
```
HeitorAugustoLN, line 23: I would reduce both to this but this is just to make it shorter, it doesn't change anything

```suggestion
    xdg.configFile."ghostty/themes/${themeName}".source = "${sources.ghostty}/${themeName}.conf";

    programs.ghostty.settings.theme = "light:${themeName},dark:${themeName}";
```

HeitorAugustoLN (approved):
LGTM! I have a suggestion for making the module shorter, but it doesn't really change anything besides that

getchoo:
Tested on NixOS and all seems good!

CI did catch a new issue though: our tests against NixOS/home-manager stable fail to evaluate due to the `programs.ghostty` module from home-manager not being present in their 24.11 branch. It will need to be backported upstream (similar to the `ghostty` package) or ~~slight hacks~~ a neat workaround will need to be applied here to basically ignore the whole module on 24.11

Merging will be blocked by this sadly, but I'll make sure to take care of it personally since it'll probably require a good bit of effort

Thanks for your work, @natecox. I hope to ping you again soon with a merge :)

natecox:
oof, I didn't realize that it wasn't on the 24.11 branch; living on unstable bites again I guess.

getchoo:
https://github.com/nix-community/home-manager/pull/6285

getchoo (approved):
Sorry for the delay, have been with family for the last few days ;)

CI is happy now, so I think this is ready for a merge. Thanks again for your work, @natecox and I hope to see you around!

PR #447, opened 2025-01-04
Add support for chromium browsers

HeitorAugustoLN:
[catppuccin/chrome](https://github.com/catppuccin/chrome)

### DIFF
--- a/modules/home-manager/all-modules.nix
+++ b/modules/home-manager/all-modules.nix
@@ -5,6 +5,7 @@
   ./bottom.nix
   ./btop.nix
   ./cava.nix
+  ./chrome.nix
   ./cursors.nix
   ./delta.nix
   ./dunst.nix
@@ -31,6 +32,7 @@
   ./micro.nix
   ./mpv.nix
   ./neovim.nix
+  ./obs.nix
   ./polybar.nix
   ./rio.nix
   ./rofi.nix
@@ -41,7 +43,6 @@
   ./sway.nix
   ./tmux.nix
   ./tofi.nix
-  ./obs.nix
   ./waybar.nix
   ./wlogout.nix
   ./yazi.nix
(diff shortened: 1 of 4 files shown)


getchoo (changes requested):
We don't usually do this, but given that this still one port (though used across multiple programs) I think it would be better to combine these modules into just one "chrome.nix" that defines multiple options. This will let use reuse the extensions IDs across all of them

Could look something like this

```nix
{ catppuccinLib }:
{ config, lib, ... }:

let
  cfg = config.catppuccin;

  extensionIds = {
    frappe = "olhelnoplefjdmncknfphenjclimckaf";
    latte = "jhjnalhegpceacdhbplhnakmkdliaddd";
    macchiato = "cmpdlhmnmjhihmcfnigoememnffkimlk";
    mocha = "bkkmolkhemgaeaeggcmfbghljjjoofoh";
  };
in

{
  options.catppuccin = lib.genAttrs [ "brave" "chromium" "vivaldi" ] (
    name: catppuccinLib.mkCatppuccinOption { inherit name; }
  );

  config = lib.mkMerge [
    (lib.mkIf cfg.brave.enable {
      programs.brave.extensions = [ { id = identifiers.${cfg.brave.flavor}; } ];
    })

    (lib.mkIf cfg.chromium.enable {
      programs.chromium.extensions = [ { id = identifiers.${cfg.chromium.flavor}; } ];
    })

    (lib.mkIf cfg.vivaldi.enable {
      programs.vivaldi.extensions = [ { id = identifiers.${cfg.vivaldi.flavor}; } ];
    })
  ];
}
```

getchoo (changes requested):
Module LGTM

Last thing to do would be enabling Chromium in `modules/tests/home.nix` to have this tested in CI, and then *probably* disabling it for macOS in `modules/tests/darwin.nix` with `lib.mkVMOverride false` (since IIRC Chromium in Nixpkgs doesn't support Mac)

HeitorAugustoLN:
Enabled chromium, vivaldi and brave in tests. Brave also works on darwin, so just needed to disable vivaldi and chromium

getchoo (approved):
Thanks!

PR #448, opened 2025-01-04
fix(home-manager/zed): correctly enable & select accent

isabelroses:
cc: @uncenter as the person who reported the accent issue to me

### DIFF
--- a/modules/home-manager/zed-editor.nix
+++ b/modules/home-manager/zed-editor.nix
@@ -7,7 +7,7 @@ let
   cfg = config.catppuccin.zed;
   enable = cfg.enable && config.programs.zed-editor.enable;
 
-  accent = if cfg.accent == "muave" then "" else " (${cfg.accent})";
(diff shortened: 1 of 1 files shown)


modules/home-manager/zed-editor.nix
```diff
@@ -5,6 +5,9 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.zed;
+  enable = cfg.enable && config.programs.zed-editor.enable;
+
+  accent = if cfg.accent == "muave" then "" else " (${cfg.accent})";
```
getchoo, line 10:
```suggestion
  accent = if cfg.accent == "mauve" then "" else " (${cfg.accent})";
```

getchoo (changes requested):
Otherwise LGTM

getchoo (approved):
Thanks!

PR #449, opened 2025-01-04
home-manager: remove IFD from helix and yazi

isabelroses:
Removes IFD. This shouldn't break themes. but if someone was previously using some form of `lib.mkForce` on these settings they will break.

### DIFF
--- a/modules/home-manager/yazi.nix
+++ b/modules/home-manager/yazi.nix
@@ -25,11 +25,10 @@ in
   };
 
   config = lib.mkIf enable {
-    programs.yazi = {
-      theme = lib.importTOML "${sources.yazi}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.toml";
-    };
-
     xdg.configFile = {
+      "yazi/theme.toml".source =
+        "${sources.yazi}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.toml";
+
       "yazi/Catppuccin-${cfg.flavor}.tmTheme".source =
         "${sources.bat}/Catppuccin ${catppuccinLib.mkUpper cfg.flavor}.tmTheme";
     };
--- a/modules/home-manager/helix.nix
+++ b/modules/home-manager/helix.nix
@@ -5,6 +5,7 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.helix;
+  enable = cfg.enable && config.programs.helix.enable;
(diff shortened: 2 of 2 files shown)


modules/home-manager/yazi.nix
getchoo, line 1: Seems this might have accidentally been pushed with this branch. I'd be fine with it in this PR, but make sure to rename the PR something like "home-manager: remove IFD from helix and yazi"

modules/home-manager/yazi.nix
```diff
@@ -25,11 +25,10 @@ in
   };
 
   config = lib.mkIf enable {
-    programs.yazi = {
-      theme = lib.importTOML "${sources.yazi}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.toml";
```
getchoo, line 29: Is there also a way we can set the theme? I assume we still need to tell yazi which one to use after linking the file to it's config dir

getchoo (changes requested):
Helix changes LGTM. Thanks!

modules/home-manager/yazi.nix
uncenter: No need, the location of the file at $XDG_CONFIG_HOME/yazi/theme.toml is all that is needed.

getchoo: Ah, brainfart. Didn't notice it was using `theme.toml` as the file name

modules/home-manager/yazi.nix
isabelroses: I did intended to push both. I just forgot to change the title 😓

modules/home-manager/yazi.nix
```diff
@@ -25,11 +25,10 @@ in
...
-    };
-
     xdg.configFile = {
+      "yazi/theme.toml".source =
+        "${sources.yazi}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.toml";
+
       "yazi/Catppuccin-${cfg.flavor}.tmTheme".source =
         "${sources.bat}/Catppuccin ${catppuccinLib.mkUpper cfg.flavor}.tmTheme";
```
getchoo, line 33: Why are we still linking this if we're using theme.toml?
It'd also probably be better to link one of these to the other, in order to avoid rewriting both on each source update

isabelroses: the `.tmTheme` file is different since that's for the syntax highlighting. but the `theme.toml` is for everything else yazi does like the selected folder.

getchoo (approved):
Thanks!

PR #450, opened 2025-01-04
fix(home-manager/alacritty): remove the `general` setting option

mariovagomarzal:
## Summary
This pull request fixes an issue in the Home-Manager Alacritty module related to the placement of the `import` config key in the generated `alacritty.toml` configuration file.

## Problem
The current implementation of the Alacritty module defines the `import` config key as follows:
```nix
settings.general.import = lib.mkBefore [ "${sources.alacritty}/catppuccin-${cfg.flavor}.toml" ];
```
This results in a configuration file with the following structure:
```toml
[general]
import = ["..."]
```
However, according to Alacritty's documentation, global settings like import must be declared at the top level of the TOML file. They should not be nested under a [general] section. The correct structure should be:
```toml
import = ["..."]

[other_options]
...
```

## Solution
This pull request modifies the module to define the `import` config key at the top level by replacing
`settings.general.import` with `settings.import`.

### DIFF
--- a/modules/home-manager/alacritty.nix
+++ b/modules/home-manager/alacritty.nix
@@ -20,7 +20,7 @@ in
 
   config = lib.mkIf cfg.enable {
     programs.alacritty = {
-      settings.general.import = lib.mkBefore [ "${sources.alacritty}/catppuccin-${cfg.flavor}.toml" ];
+      settings.import = lib.mkBefore [ "${sources.alacritty}/catppuccin-${cfg.flavor}.toml" ];
     };
   };
 }


isabelroses (approved):
Thank you :)

Yakkhini:
Wait maybe this PR has something wrong. In [latest version of alacritty's documentation](https://alacritty.org/releases/0.14.0/config-alacritty.html#general) it says the general option should under `[general]`. The global and root config is [previous version of alacritty's convention](https://alacritty.org/releases/0.13.2/config-alacritty.html#general).

Maybe you're using 0.13.2 version? I think we should follow the new api when there is conflict.

In my machine alacritty claim a warning after recent update.

yakkhini@20250105125207


Ref:
https://github.com/alacritty/alacritty/pull/8192
https://github.com/alacritty/alacritty/issues/8305

mariovagomarzal:
You're absolutely right. I seem to have come across the documentation through a link shared in a discussion on this topic and assumed it was the latest version. My bad...

PR #451, opened 2025-01-05
feat(home-manager): add support for firefox-based browsers

Anomalocaridid:
This adds support for Firefox-based browsers that have settings implemented by home manager's internal `mkFirefoxModule` function. This should also make it easy to add support for other Firefox-based browsers in the future, assuming there are no special cases that need to be handled.

Currently, this only implements support for the three web browsers I saw that use `mkFireFoxModule`, namely:
- Firefox
- Floorp
- Librewolf

I was not sure how to best share the common logic between the browsers, so I put them all in one file, `firefox.nix`. The common logic seemed like it would be too long to include in `catppuccinLib` and currently the only differences between the browsers' modules are the name of the browser itself.

Also, manually enabling the theme is needed in some cases, such as when switching between Catppuccin themes and in Floorp's case, it is needed even on a clean setup for some reason. Unfortunately, I am not sure if there is a way to address this.

Furthermore, in case it is helpful, I had to use the `old` tag from `catppuccin/firefox` because otherwise this would need to access the GitHub release artifacts, although those are the same files anyways. I assumed this would be fine because the GTK module appears to do something similar.


modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,137 @@
...
+            # Only package desired theme addon to not pollute themes list
+            # and to avoid flooding the user with extension added notifications
+            # Code adapted from `buildFirefoxXpiAddon` from
+            # https://github.com/nix-community/nur-combined/blob/master/repos/rycee/pkgs/firefox-addons/default.nix
+            (pkgs.runCommand "catppuccin-firefox" { } ''
+              dst="$out/share/mozilla/extensions/{ec8030f7-c20a-464f-9b0e-13a3a9e97384}"
+              mkdir -p "$dst"
+              install -v -m644 "${sources.firefox}/catppuccin_${cfg.flavor}_${cfg.accent}.xpi" "$dst/${addonId}.xpi"
+            '')
+          ];
```
HeitorAugustoLN, line 127: I think an option for defining the profiles to install it would be nice, instead of just assuming the profile

modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,137 @@
...
+          profiles.default.extensions = [
+            # Only package desired theme addon to not pollute themes list
+            # and to avoid flooding the user with extension added notifications
+            # Code adapted from `buildFirefoxXpiAddon` from
+            # https://github.com/nix-community/nur-combined/blob/master/repos/rycee/pkgs/firefox-addons/default.nix
+            (pkgs.runCommand "catppuccin-firefox" { } ''
+              dst="$out/share/mozilla/extensions/{ec8030f7-c20a-464f-9b0e-13a3a9e97384}"
+              mkdir -p "$dst"
+              install -v -m644 "${sources.firefox}/catppuccin_${cfg.flavor}_${cfg.accent}.xpi" "$dst/${addonId}.xpi"
+            '')
```
getchoo, line 126: I think it would be better to move this to the package rather than here. Maybe we can use split outputs? Or a scope?

modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,137 @@
...
+  pkgs,
+  ...
+}:
+
+let
+  inherit (config.catppuccin) sources;
+
+  # Each Firefox addon has an ID used to refer to it in the config
+  # Unfortunately, these addons just use randomly generated ones
+  addonIds = {
```
getchoo, line 14: Not a review, I'm just very impressed (and sorry) that you sorted through all of these. Thanks!

modules/home-manager/firefox.nix
Anomalocaridid: I have now added a `profiles` option that takes a list of profiles, defaulting with `[ "default" ]`.

modules/home-manager/firefox.nix
Anomalocaridid: I was able to move it to the package using split outputs. I have never needed to use split outputs before, so I am not sure if the way I did it is the best.

I made an output for each addon in the form `${flavor}_${accent}`. (e.g. `mocha_mauve`). I could not leave out `out` without causing an error, so I kept it. I was not sure what to put in it, so `$out` is just an empty directory.

modules/home-manager/firefox.nix
Anomalocaridid: My pleasure! I am more than happy to go to such lengths for declarative configuration.

modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,67 @@
...
+            type = lib.types.package;
+            readOnly = true;
+            description = "Resulting ${upperName} package";
+          };
+          profiles = lib.mkOption {
+            type = with lib.types; listOf str;
+            description = "${upperName} profiles to install the Catppuccin theme addon on";
+            default = [ "default" ];
+            example = lib.literalExpression ''[ "default" "default-release" "work" ]'';
+          };
```
HeitorAugustoLN, line 38: I don't really think there should be a default value in here.

```suggestion
          profiles = lib.mkOption {
            type = with lib.types; listOf str;
            description = "${upperName} profiles to install the Catppuccin theme addon on";
            example = lib.literalExpression ''[ "default" "default-release" "work" ]'';
          };
```

modules/home-manager/firefox.nix
HeitorAugustoLN: I didn't see your comment here. I don't think there should be a default value, also, the config needs update to install in all profiles, it is currently just installing on the `default` profile still

modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,67 @@
...
+                // Immutably disable syncing the active theme to avoid issues just in case
+                lockPref("services.sync.prefs.sync.extensions.activeThemeID", false);
+              '';
+            })
+          ];
+        });
+        programs.${name} = {
+          profiles.default.extensions = [
+            sources.firefox."${cfg.flavor}_${cfg.accent}"
+          ];
```
HeitorAugustoLN, line 57: Implementation should look like something like this:

```suggestion
          profiles = lib.genAttrs cfg.profiles (_: { extensions = [ sources.firefox."${cfg.flavor}_${cfg.accent}" ]; })
```

modules/home-manager/firefox.nix
Anomalocaridid: Why not? Is it not the case that everyone starts out with a profile named "default" unless they add a new one?

modules/home-manager/firefox.nix
Anomalocaridid: >the config needs update to install in all profiles, it is currently just installing on the default profile still

My bad, thank you for catching that.

pkgs/firefox/package.nix
```diff
@@ -0,0 +1,129 @@
+{
+  buildCatppuccinPort,
+  fetchFromGitHub,
+  lib,
+}:
+buildCatppuccinPort (finalAttrs: rec {
```
HeitorAugustoLN, line 6: Why make it a `rec` attribute set if there is `finalAttrs`?

modules/home-manager/firefox.nix
HeitorAugustoLN: I don't think so, for example, I use my own name for my default profile.

Doing a quick seach through github search, you can see many people don't use `default` as a profile:

https://github.com/ksevelyar/idempotent-desktop/blob/04d0508ffa28dea3e57ab23e54414f722d64551f/packages/firefox.nix#L11

https://github.com/skogsbrus/os/blob/ba9c523d8b3b086b54a1a644b1df779dbb39e1c2/home/firefox.nix#L16

https://github.com/icyphox/dotfiles/blob/76fda9709b55a6e0a6ec775c307d7bec52fb28db/programs/firefox.nix#L9

You can see more in here:

https://github.com/search?q=profiles+AND+programs+AND+firefox+language%3ANix&type=code

HeitorAugustoLN: There are plenty of people that use `default` as a profile. But this should be up to the user to configure anyways, since this is not a default that fits everyone

pkgs/firefox/package.nix
Anomalocaridid: I needed to access `addonIds` to make `outputs` and referencing it through `finalAttrs` there caused infinite recursion. I also assumed it would be better to use `finalAttrs` where I could rather than just using `rec` for everything.

Is there a better way to accomplish this?

modules/home-manager/firefox.nix
Anomalocaridid: I meant that if no profiles were configured, then `default` would be the name of the profile.

In any case, it occurred to me that most people using this would explicitly configure their profile in their config anyways, so I went ahead and added your suggestion.

pkgs/firefox/package.nix
HeitorAugustoLN: Yeah, that makes sense. I think you can just put addonIds in a `let in` and then inherit it in `passthru` then you can remove `rec` and use it to make outputs using the variable. I think this is a better way

Anomalocaridid: Good idea. I also removed `finalAttrs` because that was the only thing it was used for.

modules/home-manager/firefox.nix
```diff
@@ -0,0 +1,66 @@
...
+    {
+      options.catppuccin.${name} =
+        catppuccinLib.mkCatppuccinOption {
+          inherit name;
+          accentSupport = true;
+        }
+        // {
+          finalPackage = lib.mkOption {
+            type = lib.types.package;
+            readOnly = true;
```
isabelroses, line 30: perhpas we should mark this as `visible = false` such that the search won't have this come up?
```suggestion
            readOnly = true;
            visible = false;
```

Anomalocaridid: Done, although I added this option mainly so that users could still override the firefox-based browsers' wrappers' arguments.

bbigras:
Note that there's https://github.com/nix-community/home-manager/commit/27ffa35178bc6f359293a8809d32c5da23aaabc7 now.

I'm able to use `extensions.settings."FirefoxColor@mozilla.com".settings = builtins.fromJSON (builtins.readFile ./firefox-color.json);`

bbigras:
If I understand correctly, the new home-manager Firefox extension setting forces Firefox to store the extension settings as JSON file (instead of indexdb).

> However, would it necessitate packaging Firefox Color in this flake?

I just tested, and it's fine to create the `~/.mozilla/firefox/default/browser-extension-data/FirefoxColor@mozilla.com/storage.js` file even if the extension is not installed. If you install the extension, the file is read. Not sure what happens if the extension has a bunch of new settings to store.

> Also, where is firefox-color.json from?

I just took `~/.mozilla/firefox/default/browser-extension-data/FirefoxColor@mozilla.com/storage.js`

One thing to note is that, with the new home-manager setting, I get some `Existing file` errors. Maybe I should just use `home-manager.backupFileExtension` but it's not working in my nix-config for some reason.

While I like the new home-manager setting, I can't use it for ublock since I often need to add exceptions and I don't want them overwritten.

Anomalocaridid:
> I just tested, and it's fine to create the `~/.mozilla/firefox/default/browser-extension-data/FirefoxColor@mozilla.com/storage.js` file even if the extension is not installed. If you install the extension, the file is read. Not sure what happens if the extension has a bunch of new settings to store.

Thanks for testing it! So this approach could be viable even if users have to manually install Firefox Color themselves.

> I just took `~/.mozilla/firefox/default/browser-extension-data/FirefoxColor@mozilla.com/storage.js`

Using that directly would probably not be a good idea. However, we could use the palette to manually set the colors to have an equivalent effect like how the tty and fzf ports are implemented.

> One thing to note is that, with the new home-manager setting, I get some `Existing file` errors. Maybe I should just use `home-manager.backupFileExtension` but it's not working in my nix-config for some reason.

Somewhat off-topic, but I think `home-manager.backupFileExtension` needs to be set in the current generation before it has any effect.

With all that said, I think this port is blocked on a few questions that I need help in answering:
- Should we use the "legacy" themes or switch to Firefox Color?
    - The "legacy" themes are potentially at risk of becoming outdated at some point since they technically aren't being maintained (i.e. if something they rely on gets deprecated or removed from Firefox), but this probably will not be a problem for a while. However, we would not have to worry about how to handle dependencies on non-Catppuccin Firefox extensions.
    - Firefox Color would result in another module depending on the palette and having to manually sync colors if the themes change going forward, like we currently do with the fzf and tty ports. However, it would also give users the opportunity to override any colors they want. In this case, we would probably need to do one of two things:
        - Make users install Firefox Color themselves. (Which would be slightly inconvenient)
        - Package Firefox Color ourselves. (Which may or may not be out of scope)
    - Since it is now feasible to support other extensions like, say, Dark Reader and Stylus, if we want to eventually do so, we would still need to figure out how to handle reliance on non-Catppuccin Firefox extensions anyways.
- Since the changes to `firefox.extensions` will need to be supported regardless of what approach we use, would it need to be polyfilled for older versions of Home Manager? If so, how?
    - This may not be a concern if we go the Firefox Color route since we would have to rely on the changes anyways.

getchoo:
Taking this off the milestone since it seems this will take a good bit of effort still

Anomalocaridid:
Closing in favor of #592

PR #452, opened 2025-01-05
revert: "fix(home-manager/alacritty): remove the `general` setting option"

Yakkhini:
Reverts catppuccin/nix#450

https://github.com/catppuccin/nix/pull/450#issuecomment-2571501256


getchoo (approved):
Thanks for catching this

As said in your comment, this is the configuration schema the current version of Alacritty (including in both nixpkgs unstable and 24.11) expects; It's better to go with that

PR #454, opened 2025-01-05
feat(home-manager): add support for lsd

mariovagomarzal:
Add support for the LSD program.

In particular, the module defines the `programs.lsd.colors` option to be correspondent `themes/catppuccin-<flavor>/colors.yaml` file of the [LSD port](https://github.com/catppuccin/lsd).

### DIFF
--- a/modules/home-manager/lsd.nix
+++ b/modules/home-manager/lsd.nix
@@ -19,6 +19,9 @@ in
   };
 
   config = lib.mkIf cfg.enable {
-    programs.lsd.colors = catppuccinLib.importYAML "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
+    xdg.configFile = {
+      "lsd/config.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
+    };
+    programs.lsd.settings.color.theme = "custom";
   };
 }
--- a/modules/home-manager/lsd.nix
+++ b/modules/home-manager/lsd.nix
@@ -13,7 +13,7 @@ in
 
   config = lib.mkIf enable {
     xdg.configFile = {
-      "lsd/config.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
+      "lsd/colors.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
     };
     programs.lsd.settings.color.theme = "custom";
   };
--- a/modules/home-manager/lsd.nix
+++ b/modules/home-manager/lsd.nix
@@ -1,0 +1,24 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.catppuccin.lsd;
+in
+
+{
+  options.catppuccin.lsd = catppuccinLib.mkCatppuccinOption { name = "lsd"; };
+
+  imports = catppuccinLib.mkRenamedCatppuccinOptions {
+    from = [
+      "programs"
+      "lsd"
+      "catppuccin"
+    ];
+    to = "lsd";
+  };
+
+  config = lib.mkIf cfg.enable {
+    programs.lsd.colors = catppuccinLib.importYAML "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
+  };
+}
--- a/modules/home-manager/lsd.nix
+++ b/modules/home-manager/lsd.nix
@@ -3,7 +3,9 @@
 
 let
   inherit (config.catppuccin) sources;
+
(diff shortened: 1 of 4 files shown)


modules/home-manager/lsd.nix
```diff
@@ -0,0 +1,24 @@
...
+    from = [
+      "programs"
+      "lsd"
+      "catppuccin"
+    ];
+    to = "lsd";
+  };
+
+  config = lib.mkIf cfg.enable {
+    programs.lsd.colors = catppuccinLib.importYAML "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
```
uncenter, line 22:
```suggestion
    xdg.configFile = {
      "lsd/colors.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
    }
```

I think you can do this to avoid IFD. https://github.com/lsd-rs/lsd#color-theme

mariovagomarzal: That was my initial thought as well. However, upon discovering the [`programs.lsd.colors`](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.lsd.colors) option, I decided it would be more convenient to leverage it for the implementation.

In any case, with your suggested approach, do you think we should explicitly set
```nix
programs.lsd.settings.color.theme = "custom";
```
to ensure LSD uses the colors.yaml file? Note that this step is handled automatically when using the `lsd.colors` option.

isabelroses: I would prefer to remove IFD with the `colours.yml` file. we should also apply the theme as you suggested.

modules/home-manager/lsd.nix
```diff
@@ -0,0 +1,27 @@
...
+  imports = catppuccinLib.mkRenamedCatppuccinOptions {
+    from = [
+      "programs"
+      "lsd"
+      "catppuccin"
+    ];
+    to = "lsd";
+  };
+
+  config = lib.mkIf cfg.enable {
```
isabelroses, line 21:
```suggestion
  config = lib.mkIf enable {
```

modules/home-manager/lsd.nix
mariovagomarzal: I've committed the changes. Let me know if it is to your liking.

modules/home-manager/lsd.nix
```diff
@@ -0,0 +1,27 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.catppuccin.lsd;
```
isabelroses, line 6: we need to include this new enable, because we don't want to install the theme to someone's system who isn't expecting it. This previously would have been gated by `config.programs.lsd.enable` anyways prior to moving to `xdg.configFile`
```suggestion

  cfg = config.catppuccin.lsd;
  enable = cfg.enable && config.programs.lsd.enable;
```

isabelroses (changes requested):

modules/home-manager/lsd.nix
```diff
@@ -0,0 +1,29 @@
...
+
+  imports = catppuccinLib.mkRenamedCatppuccinOptions {
+    from = [
+      "programs"
+      "lsd"
+      "catppuccin"
+    ];
+    to = "lsd";
+  };
+
```
isabelroses, line 22: We also don't need this code because it was for the moving to the catppuccin name space, but this port is already under the namespace.
```suggestion

```

mariovagomarzal: Okay, sorry, I don't understand this.

isabelroses (approved):

mariovagomarzal:
Why is the "CI / Build Packages (macos-13)" check taking so long and failing?

modules/home-manager/lsd.nix
```diff
@@ -0,0 +1,20 @@
...
+  cfg = config.catppuccin.lsd;
+  enable = cfg.enable && config.programs.lsd.enable;
+in
+
+{
+  options.catppuccin.lsd = catppuccinLib.mkCatppuccinOption { name = "lsd"; };
+
+  config = lib.mkIf enable {
+    xdg.configFile = {
+      "lsd/config.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
```
isabelroses, line 16: Just noticed this should be `colors.yaml` not `config.yaml`
```suggestion
      "lsd/colors.yaml".source = "${sources.lsd}/catppuccin-${cfg.flavor}/colors.yaml";
```

isabelroses (approved):

isabelroses:
I will merge this after CI finishes. The mac modules check can be a bit slow, so if it stalls i'll check on my mac.

Issue #259, opened 2024-06-30
transparent_background and other options for catppuccin/nvim


reinux:
There's a slew of options for [catppuccin/nvim](https://github.com/catppuccin/nvim) that might be worth being able to set.

On the other hand, the other apps catppuccin configures don't seem to have so many options made available, so I'm not sure if we want to do so many. (Cava does have a transparency option.)

I did make [a fork](https://github.com/reinux/catppuccin-nix) with the transparent_background option, but I don't want to make a PR just yet, because this is my first time writing options for any nix packages, and I'm not confident regarding style and conventions etc. (e.g. do we keep the snake case?)

Any guidance would be appreciated!

getchoo:
a [freeform module](https://nixos.org/manual/nixos/stable/#sec-freeform-modules) would probably be ideal here as we wouldn't need to constantly update our options to reflect the upstream port

PRs are welcome here

toodeluna:
I made a pull request for this that adds a new option `extraConfig` which gets dynamically converted to a Lua table. It does use the type attrs instead of a freeform module which (as far as I can see) makes it easier to create the plugin settings dynamically.

PR #457, opened 2025-01-12
feat(home-manager/neovim): add option for custom settings

toodeluna:
This commit adds an `extraConfig` option to the neovim options that gets dynamically converted into a Lua table and passed to the setup function which is useful for configuring things like italics, transparent_background,...

Fixes this issue: https://github.com/catppuccin/nix/issues/259

### DIFF
--- a/modules/home-manager/neovim.nix
+++ b/modules/home-manager/neovim.nix
@@ -7,10 +7,38 @@
 
 let
   cfg = config.catppuccin.nvim;
+
+  toLuaValue =
+    value:
+    if builtins.isAttrs value then
+      toLuaTable value
+    else if lib.isString value || lib.isPath value then
+      ''"${toString value}"''
+    else if lib.isList value then
+      "{${lib.strings.concatStringsSep "," (map toLuaValue value)}}"
+    else if lib.isBool value then
+      if value == true then "true" else "false"
+    else
+      toString value;
+
+  toLuaTable =
+    attrset:
+    let
+      values = lib.strings.concatStringsSep "," (
+        lib.attrsets.mapAttrsToList (name: value: "${name} = ${toLuaValue value}") attrset
+      );
+    in
+    "{${values}}";
 in
 
 {
-  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; };
+  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; } // {
+    extraConfig = lib.mkOption {
+      description = "Extra settings that will be passed to the setup function.";
+      default = { };
+      type = lib.types.attrs;
+    };
+  };
 
   imports = catppuccinLib.mkRenamedCatppuccinOptions {
     from = [
@@ -32,10 +60,18 @@ in
               vim.fn.mkdir(compile_path, "p")
               vim.opt.runtimepath:append(compile_path)
 
-              require("catppuccin").setup({
-              	compile_path = compile_path,
-              	flavour = "${cfg.flavor}",
-              })
+              local catppuccin_custom_plugin_settings = ${toLuaTable cfg.extraConfig}
+
+              local catppuccin_plugin_settings = {
+                compile_path = compile_path,
+                flavour = "${cfg.flavor}",
+              }
+
+              for key, value in pairs(catppuccin_custom_plugin_settings) do
+                catppuccin_plugin_settings[key] = value
+              end
+
+              require("catppuccin").setup(catppuccin_plugin_settings)
 
               vim.api.nvim_command("colorscheme catppuccin")
             EOF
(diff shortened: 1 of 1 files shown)


modules/home-manager/neovim.nix
```diff
@@ -7,10 +7,38 @@
...
+      toString value;
+
+  toLuaTable =
+    attrset:
+    let
+      values = lib.strings.concatStringsSep "," (
+        lib.attrsets.mapAttrsToList (name: value: "${name} = ${toLuaValue value}") attrset
+      );
+    in
+    "{${values}}";
```
isabelroses, line 31: perhaps we should consider [`lib.generators.toLua`](https://noogle.dev/f/lib/generators/toLua)

modules/home-manager/neovim.nix
```diff
@@ -7,10 +7,38 @@
...
 in
 
 {
-  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; };
+  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; } // {
+    extraConfig = lib.mkOption {
+      description = "Extra settings that will be passed to the setup function.";
+      default = { };
+      type = lib.types.attrs;
```
isabelroses, line 39: if we do use the lua generator we can do something like this too
```suggestion
      type = lib.types.submodule { freeformType = lib.types.attrsOf lib.types.anything; };
```

modules/home-manager/neovim.nix
toodeluna: Didn't realize this was a thing, but this would definitely be better than the current solution! I'll implement it to use this.

modules/home-manager/neovim.nix
toodeluna: Is there a difference between this and just using attrs?

getchoo: Currently there would be no difference, *but* I would still encourage the use of freeform modules

From the [NixOS manual](https://nixos.org/manual/nixos/unstable/#sec-freeform-modules):

> Freeform modules allow you to define values for option paths that have not been declared explicitly. This can be used to add attribute-specific types to what would otherwise have to be attrsOf options in order to accept all attribute names.

Further, this would also allow us to specify common options (like a transparent background) cleanly in our documentation rather than relying completely on the port's docs

modules/home-manager/neovim.nix
```diff
@@ -1,16 +1,22 @@
...
 
+  defaultConfig = {
+    compile_path = lib.generators.mkLuaInline "compile_path";
+    flavour = cfg.flavor;
+  };
+in
 {
-  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; };
+  options.catppuccin.nvim = catppuccinLib.mkCatppuccinOption { name = "neovim"; } // {
+    extraConfig = lib.mkOption {
```
getchoo, line 14:
```suggestion
    settings = lib.mkOption {
```

`extraConfig` usually implies a strongly typed set of options. What we're doing here is more akin to [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md), which encourages the use of `settings` for freeform options

modules/home-manager/neovim.nix
```diff
@@ -1,16 +1,22 @@
...
+      description = "Extra settings that will be passed to the setup function.";
+      default = { };
+      type = lib.types.attrs;
```
getchoo, line 17:
```suggestion
      type = lib.types.attrs;
      apply = lib.recursiveUpdate defaultConfig;
```
It'd be better to apply this to the option itself so that the actual final value of it can be easily evaluated (as opposed to currently where the final configuration used is limited to the above `let ...; in` scope)

modules/home-manager/neovim.nix
```diff
@@ -1,16 +1,22 @@
...
-  config,
-  lib,
-  ...
-}:
-
+{ config, lib, ... }:
 let
   cfg = config.catppuccin.nvim;
-in
+  pluginSettings = lib.generators.toLua { } (defaultConfig // cfg.extraConfig);
```
getchoo, line 5:
```suggestion
```

modules/home-manager/neovim.nix
```diff
@@ -32,10 +38,7 @@ in
               vim.fn.mkdir(compile_path, "p")
               vim.opt.runtimepath:append(compile_path)
 
-              require("catppuccin").setup({
-              	compile_path = compile_path,
-              	flavour = "${cfg.flavor}",
-              })
+              require("catppuccin").setup(${pluginSettings})
```
getchoo, line 41:
```suggestion
              require("catppuccin").setup(${cfg.settings})
```

isabelroses (approved):

toodeluna:
@getchoo Hey, I don't want to rush you, but could these changes be approved so they can be merged into the main branch? 🙂

isabelroses (approved):

PR #458, opened 2025-01-17
refactor: move development inputs/outputs back to dev flake

getchoo:
Follow up on https://github.com/catppuccin/nix/pull/390

Now that Nix 2.26 can handle subflakes in a *much* better way, I think we're good to adopt this again. Only issue is that the subflake will **require** that in-development version of 2.26; shouldn't be a big deal though as most of the time this is only used by CI, and `nix shell` makes things easy


getchoo:
Looks like nix-eval-jobs doesn't support Nix 2.26 just yet, and I'm still having some trouble convincing the Nix installer actions to use the random build from Hydra

This will be on the back burner for now, but I consider it a blocker for 2.0 (as I'd rather not make a stable release with the Flake inputs in their current form)

isabelroses:
> I'm still having some trouble convincing the Nix installer actions to use the random build from Hydra

That's because DeterminateSystems/nix-installer-action expects "a URL pointing to the nix-installer binary". At this time I don't believe there is a way to change the nix version that is installed without specifying the version of the nix-installer or forking the repo to provide the chosen release.

getchoo:
> That's because DeterminateSystems/nix-installer-action expects "a URL pointing to the nix-installer binary".

That's for `source-url`. `nix-package-url` is what we're looking for, and it works...but only on Linux. On macOS the tarball seems to be downloaded just fine, but running commands from it fails [here](https://github.com/catppuccin/nix/actions/runs/12843158417/job/35814842743#step:3:44). Still an improvement over the situation with the Cachix though, where we would need to use the official installation script -- which even when built on Hydra links to a (non-existent) tarball on releases.nixos.org 🫠 

Might need to just use a different revision of Nix from Hydra to fix the Mac issues, but not sure yet

getchoo:
> Looks like nix-eval-jobs doesn't support Nix 2.26 just yet

Pulled in https://github.com/nix-community/nix-eval-jobs/pull/341, which adds support for the new Meson infra in Nix. It fixes building nix-eval-jobs/nix-fast-build, but evaluation still fails; might be a regression in the Nix build. Will need to play with it more

If I can figure out the macOS installation issues first, we may just switch to a more traditional CI scheme w/o nix-fast-build. Would probably be a simple `nix flake check` to build the module tests and a `linkFarm` of all of our packages

getchoo:
Nix 2.26 released and I implemented the CI changes mentioned above. This should be good to go!

isabelroses (approved):

PR #465, opened 2025-02-04
chore: set module class

isabelroses:
this *should* help prevent errors from utilising the wrong class for the diffrent system types. e.g. darwin trying to use nixos modules

### DIFF
--- a/modules/home-manager/default.nix
+++ b/modules/home-manager/default.nix
@@ -1,6 +1,8 @@
 { lib, ... }:
 
 {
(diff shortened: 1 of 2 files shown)


getchoo (changes requested):
This should be placed in the default.nix file of each module set (i.e., `modules/nixos/default.nix`) so that it's applied across both Flake and stable Nix configurations

getchoo (approved):
Thanks!

PR #469, opened 2025-02-08
fix(whiskers): enable useFetchCargoVendor

PerchunPak:
Using `buildRustPackage` without it is deprecated
https://github.com/NixOS/nixpkgs/pull/378288#issuecomment-2644744513


PerchunPak:
weird CI...

PerchunPak:
Without this, it also throws a hash mismatch if you are on newer nixpkgs

isabelroses:
Thanks for the fix.

PR #477, opened 2025-02-26
feat(darwin): init

### LINKED ISSUE #162: Support Nix Darwin
Would it be something possible to add a module to support [nix-darwin](https://github.com/LnL7/nix-darwin) system configurations ?

Trying to use catppuccin-nix in a darwin module cause this error (grub isn't a thing in nix-darwin):
```
error: The option `boot.loader.grub.font' does not exist. Definition values:
       - In `<unknown-file>':
           {
             _type = "if";
             condition = false;
             content = "/nix/store/9dgvl02bvzsadmwsq014r14509bf6fmz-catppuccin-grub-theme/font.pf2";
           }
```

isabelroses:
previously it made no sense to add this since we didn't support any darwin modules but now we have the global `cache.enable` and some users may want to set it in a "shared" module between nixos and darwin

closes https://github.com/catppuccin/nix/issues/162

### DIFF
--- a/dev/flake.lock
+++ b/dev/flake.lock
@@ -39,16 +39,36 @@
         ]
       },
       "locked": {
-        "lastModified": 1779336838,
-        "narHash": "sha256-n1+l78hJRABp4cQHKeD0BVByT0vZLPqd09Tvoq8Q+d8=",
+        "lastModified": 1779726696,
+        "narHash": "sha256-/p37CB5n6Wpw250b0Lq0CYwNq2D8uGKzDoBulyLcQqA=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "928d72376949e222ea4f07b44828a55b0136422e",
+        "rev": "1a95e2efb477959b70b4a14c51035975c0481df6",
         "type": "github"
       },
       "original": {
         "owner": "nix-community",
         "repo": "home-manager",
+        "type": "github"
+      }
+    },
+    "nix-darwin": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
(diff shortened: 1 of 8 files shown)


getchoo:
Sorry it took me a bit to review

When I first saw this in my inbox I initially just thought "no", but after mulling over it again, this would be very helpful for Darwin HM users instead of needing to configure this at a per-user level or shove it in a Flake config

The only problem I have now is the expectations that exposing this module sets. As someone taking a quick look at this project (or well, our flake.nix), I would assume we support Darwin the same as NixOS/HM. It'd be a bit weird to find out this is the only option...so I dunno

isabelroses:
> When I first saw this in my inbox I initially just thought "no"

Honestly this was my reaction last time we had one of these 😔 

> this would be very helpful for Darwin HM users instead of needing to configure this at a per-user level or shove it in a Flake config

Honestly thats the only reason it makes any sense right now since. And it somewhat makes sense to reduce that complexity when setting up but thats pretty niche.

> The only problem I have now is the expectations that exposing this module sets. As someone taking a quick look at this project (or well, our flake.nix), I would assume we support Darwin the same as NixOS/HM. 

Totally valid. I think it maybe a good idea to get a consensus on this. Perhaps in the discord?

isabelroses:
Actually might be worth merging now that https://github.com/catppuccin/nix/issues/545 is open

isabelroses:
i feel like we totally forgot about the existence of fish, tmux, and vim. when initially discussing this. i want to expand our none home-manger support out a bit more, hence https://github.com/catppuccin/nix/pull/953. so lets reconsider this one

isabelroses:
testing in ci is funny

PR #478, opened 2025-02-28
feat(home-manager): nushell init

isabelroses:


natecox:
neat, I was just coming in to start this port myself. +1.

modules/home-manager/nushell.nix
```diff
@@ -0,0 +1,20 @@
...
+
+  cfg = config.catppuccin.nushell;
+in
+
+{
+  options.catppuccin.nushell = catppuccinLib.mkCatppuccinOption { name = "nushell"; };
+
+  config = lib.mkIf cfg.enable {
+    programs.nushell = {
+      extraConfig = lib.mkAfter ''
```
AndrewCouncil, line 15: should this mirror what zsh syntax highlighting does and use `lib.mkBefore`? this would allow for user overrides i think. it looks like that is the typical way to add these import statements, as i don't find any existing modules that use `lib.mkAfter`.
https://github.com/catppuccin/nix/blob/f833a338ff6c091c84e041ee77ce88f8b242ca79/modules/home-manager/zsh-syntax-highlighting.nix#L67C1-L67C36

AndrewCouncil (approved):
incredible, excited to get rid of the jank way i am doing this is my config

modules/home-manager/nushell.nix
isabelroses: Thats a good point!

isabelroses:
I've been using this branch for the last ~5 days, seems fairly stable.

PR #481, opened 2025-03-04
ci: avoid having to approve first time contributors

isabelroses:
`pull_request` -> `pull_request_target` means that we don't need to approve first time contributors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - 'pkgs/**'
 
       - '.github/workflows/ci.yml'
-  pull_request:
+  pull_request_target:
     paths:
       - '**.lock'
       - '**.nix'
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -11,7 +11,7 @@ on:
       - 'CHANGELOG.md'
 
       - '.github/workflows/website.yml'
-  pull_request:
+  pull_request_target:
     paths:
       - '**.lock'
       - '**.nix'


sgoudham:
Let's be careful here, I'm pretty sure the target event has a lot of caveats and security considerations.

isabelroses:
> Let's be careful here, I'm pretty sure the target event has a lot of caveats and security considerations.

At worst I think a bad build could be pushed to the cachix but a user would have to intentionally pull that since it wouldn't make its way to main.

> (I also kinda want to veto this and say that all first time contributors to the repository should always have their workflows approved by a trusted person)

I understand that but its a bit of a hassle when you have to approve the same PR running 3-4 times, over relatively small issues.

sgoudham:
Perhaps a solution is to change the option to only new GitHub users in the settings if it's becoming a hassle:

image

I'm still not entirely sold since I'd still prefer first time contributors to have approvals but I won't argue with your experience if you're annoyed enough to raise a PR for it.

sgoudham:
I've updated the option in the settings to only require approvals for new GitHub users that haven't contributed to the repo @isabelroses

Can we trial that for a while?

isabelroses:
> I've updated the option in the settings to only require approvals for new GitHub users that haven't contributed to the repo @isabelroses
> 
> Can we trial that for a while?

We can seems more appropriate.

getchoo:
> all first time contributors to the repository should always have their workflows approved by a trusted person

I agree.

Moving to `pull_request_target` also allows for PR runs to have access to the repository's secrets, which isn't something I've kept in mind while writing them -- though from a quick look, I don't *think* you could do anything with them

> I've updated the option in the settings to only require approvals for new GitHub users that haven't contributed to the repo

But if you have an issue with it @isabelroses, I don't really mind this solution at all

isabelroses:
From my experience so far change the repo settings has worked just fine! Thanks.

PR #484, opened 2025-03-07
ci: fix update locks workflow

getchoo:


.github/workflows/update-locks.yml
```diff
@@ -63,9 +63,9 @@ jobs:
 
       - name: Install Nix
         uses: cachix/install-nix-action@V27
-
-      - name: Install python
-        run: nix profile install --inputs-from . 'nixpkgs#python3'
+        with:
+          # NOTE: Required for paws
+          nix_path: nixpkgs=channel:nixos-unstable
```
getchoo, line 68: @isabelroses we could also use the new [`nix` shebang intepreter](https://nix.dev/manual/nix/2.26/command-ref/new-cli/nix#shebang-interpreter). Thoughts?

isabelroses: I think purely from the point of compatibility what you proposed here is good. But I'm indifferent.

PR #492, opened 2025-03-10
refactor: add `catppuccin-` prefix to `buildCatppuccinPort` pnames

### LINKED ISSUE #491: `buildCatppuccinPort` derivation names can be confused with eponymous packages
The usage of `buildCatppuccinPort` returns derivations that have names in the format of `<port>-<rev>`; when `rev` happens to start with a number, parsers consider the package name to be `<port>` (according to the standard behavior of `builtins.parseDrvName`). This causes confusion with software packages with the same name from Nixpkgs or elsewhere for tools such as [nvd](https://khumba.net/projects/nvd).

For instance, nvd would interpret `fish-1234567` as a package with name `fish` and version `1234567`, and would display this in diffs and lists as an additional version installed alongside the `fish` software package.

ryanccn:
Closes #491.

### DIFF
--- a/pkgs/buildCatppuccinPort/package.nix
+++ b/pkgs/buildCatppuccinPort/package.nix
@@ -17,11 +17,14 @@ stdenvNoCC.mkDerivation (
 
   args'
   // {
-    version = args'.version or (builtins.substring 0 7 finalAttrs.src.rev);
+    pname = args'.pname or "catppuccin-${finalAttrs.port}";
+    version =
+      args'.version
+        or ("0" + lib.optionalString (finalAttrs ? "lastModified") "-unstable-${finalAttrs.lastModified}");
 
     src =
-      args'.src or sources.${finalAttrs.pname} or (fetchCatppuccinPort {
-        port = finalAttrs.pname;
+      args'.src or sources.${finalAttrs.port} or (fetchCatppuccinPort {
+        inherit (finalAttrs) port;
         inherit (finalAttrs) rev hash;
         fetchSubmodules = finalAttrs.fetchSubmodules or false;
       });
@@ -29,8 +32,8 @@ stdenvNoCC.mkDerivation (
     nativeBuildInputs = args'.nativeBuildInputs or [ ] ++ [ catppuccinInstallHook ];
 
     meta = {
-      description = "Soothing pastel theme for ${finalAttrs.pname}";
-      homepage = "https://github.com/catppuccin/${finalAttrs.pname}";
+      description = "Soothing pastel theme for ${finalAttrs.port}";
+      homepage = "https://github.com/catppuccin/${finalAttrs.port}";
       license = lib.licenses.mit;
       maintainers = with lib.maintainers; [
         getchoo
--- a/default.nix
+++ b/default.nix
@@ -13,13 +13,17 @@ let
     let
       generated = lib.foldlAttrs (
         acc: port:
-        { rev, hash }:
+        {
+          rev,
+          hash,
+          lastModified,
+        }:
         lib.recursiveUpdate acc {
           # Save our sources for each port
           sources.${port} = catppuccinPackages.fetchCatppuccinPort { inherit port rev hash; };
 
           # And create a default package for them
-          "${port}" = catppuccinPackages.buildCatppuccinPort { pname = port; };
+          "${port}" = catppuccinPackages.buildCatppuccinPort { inherit port lastModified; };
         }
       ) { } (lib.importJSON ./pkgs/sources.json);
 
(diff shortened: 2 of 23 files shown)


pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,21 @@ stdenvNoCC.mkDerivation (
 
   args'
   // {
-    version = args'.version or (builtins.substring 0 7 finalAttrs.src.rev);
+    pname = args'.pname or "ctp-${args'.port}";
```
isabelroses, line 20: I'd personally prefer to be more verbose with this.
```suggestion
    pname = args'.pname or "catppuccin-${args'.port}";
```

pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,21 @@ stdenvNoCC.mkDerivation (
...
+    version = args'.version or "0+${builtins.substring 0 7 finalAttrs.src.rev}";
```
isabelroses, line 21: I do believe `nvd` is intended for semver, so it would not be better to provide this structure.
```suggestion
    version = args'.version or "0.0.0+${builtins.substring 0 7 finalAttrs.src.rev}";
```

isabelroses (commented):
LGTM other then some small notes

pkgs/buildCatppuccinPort/package.nix
ryanccn: `nvd` has very lax version parsing rules, and the `0+` here merely serves to indicate the start of the version number for the benefit of tools such as `nvd` (since [versions must start with digits](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#versioning)). If the last modified dates were written into `sources.json`,  however, the more informative `"0-unstable-1970-01-01"` format could be used; would this usage justify the additional data?

isabelroses: Userstyles moved to [calver](https://calver.org/) which may actually be a good idea here too. And nvd would be able to detect that too! Though it would make the versions look funny being at `2025.02.31` for example.

pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,22 @@ stdenvNoCC.mkDerivation (
 
   args'
   // {
-    version = args'.version or (builtins.substring 0 7 finalAttrs.src.rev);
+    pname = args'.pname or "catppuccin-${args'.port}";
```
getchoo, line 20:
```suggestion
    pname = args'.pname or "catppuccin-${finalAttrs.port}";
```

pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,22 @@ stdenvNoCC.mkDerivation (
...
+    version =
+      args'.version or (if args' ? "lastModified" then "0-unstable-${args'.lastModified}" else "0");
```
getchoo, line 22:
```suggestion
      args'.version or "0" + lib.optionalString (finalAttrs ? "lastModified") "-unstable-${finalAttrs.lastModified}";
```

pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,22 @@ stdenvNoCC.mkDerivation (
...
 
     src =
-      args'.src or sources.${finalAttrs.pname} or (fetchCatppuccinPort {
-        port = finalAttrs.pname;
+      args'.src or sources.${args'.port} or (fetchCatppuccinPort {
+        inherit (args') port;
```
getchoo, line 26:
```suggestion
      args'.src or sources.${finalAttrs.port} or (fetchCatppuccinPort {
        inherit (finalAttrs) port;
```

pkgs/buildCatppuccinPort/package.nix
```diff
@@ -17,20 +17,22 @@ stdenvNoCC.mkDerivation (
...
         fetchSubmodules = finalAttrs.fetchSubmodules or false;
       });
 
     nativeBuildInputs = args'.nativeBuildInputs or [ ] ++ [ catppuccinInstallHook ];
 
     meta = {
-      description = "Soothing pastel theme for ${finalAttrs.pname}";
-      homepage = "https://github.com/catppuccin/${finalAttrs.pname}";
+      description = "Soothing pastel theme for ${args'.port}";
+      homepage = "https://github.com/catppuccin/${args'.port}";
```
getchoo, line 35:
```suggestion
      description = "Soothing pastel theme for ${finalAttrs.port}";
      homepage = "https://github.com/catppuccin/${finalAttrs.port}";
```

getchoo (commented):
Really like what you did with the versioning here as well. `nvd` will be very happy :)

getchoo (approved):
Thanks!

Issue #493, opened 2025-03-10
rofi theme does not work after update


MisileLab:
![Image](https://github.com/user-attachments/assets/73fa3726-77a2-4f30-ad60-8d8e1921b6cf)
maybe reason is https://github.com/catppuccin/rofi/pull/31.

PR #494, opened 2025-03-10
fix: rofi

isabelroses:
@MisileLab for testing

fixes https://github.com/catppuccin/nix/issues/493 ...maybe


modules/home-manager/rofi.nix
```diff
@@ -19,10 +20,17 @@ in
     to = "rofi";
   };
 
-  config = lib.mkIf cfg.enable {
+  config = lib.mkIf enable {
     programs.rofi = {
       theme = {
-        "@theme" = "${sources.rofi}/catppuccin-${cfg.flavor}.rasi";
+        "@theme" = "catppuccin-default";
```
getchoo, line 26: Doesn't this need to be linked as well? How does Rofi know about this path?

isabelroses: Ahh good call.

getchoo (approved):

Issue #496, opened 2025-03-12
Rofi has error on launch


typedrat:
```
❯ rofi -show drun
Failed to open theme: <i>/home/awilliams/catppuccin-mocha.rasinc</i>
Error: <b>No such file or directory</b>
```

This only started after the update to Rofi theming a few commits ago. It should be noted that something is doubly confused, because I'm using Frappe and not Mocha.

typedrat:
Yeah, specifically, the issue is that `catppuccin-default.rasi` always imports `catppuccin-mocha` no matter what.

isabelroses:
> Yeah, specifically, the issue is that `catppuccin-default.rasi` always imports `catppuccin-mocha` no matter what.

Good spot. I can patch that out.

PR #497, opened 2025-03-12
fix(pkgs/rofi): remove import at build

isabelroses:
@typedrat for testing

closes https://github.com/catppuccin/nix/issues/496


typedrat:
Works a treat, thanks!

PR #480, opened 2025-03-03
feat(home-manager): add support for thunderbird

nim65s:
Hi,

This PR fix #349

PR #499, opened 2025-03-13
feat(home-manager): add support for firefox

nim65s:
Hi,

This mimic #480, but for firefox.

I guess #451 is better because it has more features, but for now it is stalled, so I think I can share this quick and simple solution for those looking for one, even if this probably won't get merged.


getchoo:
I honestly would prefer a smaller implementation like this generally, but given that these extensions are no longer supported, we can't really ship them

Thanks though!

PR #500, opened 2025-03-13
fix(nixos/gitea): remove toString causing gc to remove it

isabelroses:


isabelroses:
oh it only allows string args ;o;

PR #503, opened 2025-03-14
fix(pkgs/paws): set timezone to UTC for fetcher

isabelroses:
By default [`datetime.datetimefromtimestamp`](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) uses localtime , so to prevent more "spammy" sources.json changes, like https://github.com/catppuccin/nix/pull/501, lets set the timezone to UTC.


getchoo (approved):
Thanks!

PR #535, opened 2025-04-11
feat(home-manager): add support for sioyak

n194:
Hai, I've added a module for sioyak. 
Although I'm not sure what is the best way to handle when `programs.sioyek.config.startup_commands` already has a value.

```shell
error: The option `home-manager.users.n.programs.sioyek.config.startup_commands' has conflicting definition values:
- In `/nix/store/saj8wg1a0iwxynwg2viywh57mfxqbjia-source/modules/home-manager/sioyek.nix': "toggle_custom_color;"
- In `/nix/store/6p5182xf5wbqvrysrraj9kcbm101yjsq-source/home/programs.nix': "toggle_scrollbar"
  Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

### DIFF
--- a/modules/home-manager/sioyek.nix
+++ b/modules/home-manager/sioyek.nix
@@ -12,29 +12,11 @@ let
 in
  
 {
-  options.catppuccin.sioyek = 
-    catppuccinLib.mkCatppuccinOption {
-      name = "sioyek";
-    }
-    // {
-      applyOnStartup = lib.mkOption {
-        type = lib.types.bool;
-        default = true;
-        description = ''
-          Applies the theme by adding `toggle_custom_color` to `startup_commands` in sioyek's config.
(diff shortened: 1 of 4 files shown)


modules/home-manager/sioyek.nix
```diff
@@ -0,0 +1,48 @@
...
+    };
+
+  imports = catppuccinLib.mkRenamedCatppuccinOptions {
+    from = [
+      "programs"
+      "sioyek"
+      "catppuccin"
+    ];
+    to = "sioyek";
+  };
```
isabelroses, line 37: This is for older modules before we namespaced them. so we don't need this for new modules
```suggestion
```

isabelroses (approved):

isabelroses:
Thanks for the PR. Hopefully those upstream changes come through soon. ❤️

PR #536, opened 2025-04-16
Update starship.nix to match starship.toml

phucleeuwu:
this one: https://github.com/catppuccin/starship/blob/main/starship.toml

### DIFF
--- a/modules/home-manager/starship.nix
+++ b/modules/home-manager/starship.nix
@@ -24,21 +24,21 @@ in
       settings = {
         format = lib.mkDefault "$all";
         palette = "catppuccin_${cfg.flavor}";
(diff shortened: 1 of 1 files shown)


isabelroses (commented):
From my understanding the starship.toml is supposed to show how you *could* use the theme. Which is why we don't include it by default

phucleeuwu:
> From my understanding the starship.toml is supposed to show how you _could_ use the theme. Which is why we don't include it by default

but the default picture look like this, I think the config should look like picture: 
image

uncenter:
That is only an example of how to use the theme, as @isabelroses said and as the [usage section](https://github.com/catppuccin/starship#usage) mentions. We don't offer opinionated settings such as a customized prompt setup, instead we offer the ability (via providing palettes) for you to customize it to your liking.

PR #543, opened 2025-04-22
refactor: zsh-syntax-highlighting initExtra -> initContent

khaneliman:
Will be deprecated shortly, switch to new configuration before users get warnings from it https://github.com/nix-community/home-manager/pull/6841

Using `mkOrder 950` to keep position same in generated output as before. (Top of the `initExtra` location.

### DIFF
--- a/modules/home-manager/zsh-syntax-highlighting.nix
+++ b/modules/home-manager/zsh-syntax-highlighting.nix
@@ -63,11 +63,16 @@ in
 
     # And this is our actual module
     (lib.mkIf cfg.enable {
-      programs.zsh = {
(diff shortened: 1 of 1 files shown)


khaneliman:
Looks like it requires a dev flake lock update for the unstable HM.

isabelroses:
> but this option has been available since 2025-03-13 and the dev flake is locked to 2025-04-17 home-manager.

Thats because we test for both unstable and release 24.11. The best thing to do here is check `stateVersion` or if the option new exists.

isabelroses (approved):
Thank you ❤️

PR #544, opened 2025-04-24
fix(home-manager/fcitx5): support new enable option

brian14708:
see: nix-community/home-manager#6885

### DIFF
--- a/modules/home-manager/fcitx5.nix
+++ b/modules/home-manager/fcitx5.nix
@@ -5,7 +5,16 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.fcitx5;
-  enable = cfg.enable && config.i18n.inputMethod.enabled == "fcitx5";
+  enable =
+    cfg.enable
+    && (
+      (
+        config.i18n.inputMethod ? enable
+        && config.i18n.inputMethod.enable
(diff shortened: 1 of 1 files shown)


modules/home-manager/fcitx5.nix
```diff
@@ -5,7 +5,12 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.fcitx5;
-  enable = cfg.enable && config.i18n.inputMethod.enabled == "fcitx5";
+  enable =
+    cfg.enable
+    && (
+      (config.i18n.inputMethod ? enable && config.i18n.inputMethod.type == "fcitx5")
```
isabelroses, line 11: This isn't actually checking if `config.i18n.inputMethod.enable` is true.

e.g.

```console
nix-repl> foo = { bar = false; }
Added foo.

nix-repl> foo ? bar
true

nix-repl> foo ? bar && foo.bar
false
```

```suggestion
      (config.i18n.inputMethod ? enable && config.i18n.inputMethod.enable && config.i18n.inputMethod.type == "fcitx5")
```

isabelroses (approved):

isabelroses:
Thanks for the PR

PR #563, opened 2025-05-14
fix(mako): deprecated settings.criteria

brian14708:
see: https://github.com/nix-community/home-manager/pull/7024
closes https://github.com/catppuccin/nix/issues/564

### DIFF
--- a/modules/home-manager/mako.nix
+++ b/modules/home-manager/mako.nix
@@ -38,15 +38,7 @@ in
   config.services.mako = lib.mkIf cfg.enable (
     if (config.services.mako ? settings) then
       {
-        settings = {
-          inherit (theme)
-            background-color
-            text-color
-            border-color
-            progress-color
-            ;
-        };
-        criteria = extraConfigAttrs;
+        settings = theme;
       }
     else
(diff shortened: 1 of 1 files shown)


brian14708:
there is a problem with the encoding of ini-sections before upstream deprecation.

is there a better way to handle this?

isabelroses:
I think now we can just do `settings = theme` the only reason we were inheriting before hand was due to the limitations of home-manager.

etu:
I'm not affiliated with the project what so ever, however, I've tested the commit e82a5b99135d0ea54d4ac12e595865f8b2cfee08 in my local repo and it seems to work fine :slightly_smiling_face:

isabelroses:
> I'm not affiliated with the project what so ever, however, I've tested the commit [e82a5b9](https://github.com/catppuccin/nix/commit/e82a5b99135d0ea54d4ac12e595865f8b2cfee08) in my local repo and it seems to work fine 🙂

Thanks for testing. I asked another person before this and she said it was good too!

isabelroses (approved):

PR #576, opened 2025-05-30
add option to enable pointerCursor

phucleeuwu:
https://github.com/catppuccin/cursors#nixos


isabelroses (commented):
We already set cursors here https://github.com/catppuccin/nix/blob/main/modules/home-manager/cursors.nix

PR #587, opened 2025-06-16
feat(home-manager): add support for xfce4-terminal

different-name:


different-name:
Build is failing due to `lib.hm.strings.toKebabCase` not being available in stable home-manager yet

different-name:
Build is failing due to `xfconf.settings` not being supported on darwin, I am not sure what the approach to a linux exclusive module is

modules/home-manager/xfce4-terminal.nix
```diff
@@ -0,0 +1,53 @@
...
+  cfg = config.catppuccin.xfce4-terminal;
+  enable = cfg.enable;
+  themeName = "catppuccin-${cfg.flavor}";
+
+  toCaseWithSeparator =
+    separator: string:
+    let
+      splitByWords = builtins.split "([A-Z])";
+      processWord = s: if lib.isString s then s else separator + lib.toLower (lib.elemAt s 0);
+      words = splitByWords string;
```
isabelroses, line 15: Why not use https://noogle.dev/f/lib/splitStringBy

modules/home-manager/xfce4-terminal.nix
```diff
@@ -0,0 +1,53 @@
...
+  options.catppuccin.xfce4-terminal = catppuccinLib.mkCatppuccinOption {
+    name = "xfce4-terminal";
+    useGlobalEnable = false;
+  };
+
+  config = lib.mkIf enable {
+    xfconf.settings = {
+      xfce4-terminal = terminalSettings;
+    };
+  };
```
isabelroses, line 52: To avoid IFD could we not just link the file to somewhere in `~/.local/share/xfce4/terminal/colorschemes/` and set `xfconf.settings.xfce4-terminal` to use that colorscheme somehow?

different-name: All themes do is set these values when you click on them, unfortunately - the theme colors are just copied into the config lol

modules/home-manager/xfce4-terminal.nix
isabelroses: Like
```nix
words = lib.splitStringBy (prev: curr: builtins.match "[a-z]" prev != null && builtins.match "[A-Z]" curr != null) string;
```

different-name: Thank you for that! I modified it a bit to produce the same output as the existing home-manager function, the idea is to switch to using that once it hits the stable lib

Not sure if it is worth making a PR for this upstream, since it's a fairly small change and I don't think this function is used very widely

different-name: Nevermind! splitStringBy is too new, build is failing because it can't find it - looks like it was added 3 months ago and I guess it hasn't landed in stable

isabelroses: This will probably be merged after https://github.com/catppuccin/nix/pull/590. So don't worry about it. :D

different-name:
> We merged #590, so if your happy to rebase. I think we can merge this too :D

On a roll!

isabelroses (approved):

Issue #186, opened 2024-05-21
package options


isabelroses:
It would be nice to be allowed to change the package for gtk, as an example. Since its hardcoded see https://github.com/catppuccin/nix/blob/57e8376fdb3514fb4c4eae432e6dfa4137f1a260/modules/home-manager/gtk.nix#L57

getchoo:
in the future i would much rather see this controllable via `catppuccin.sources`, similar to ports that don't use an entire package

this is related to an idea winston once mentioned [here](https://github.com/catppuccin/nix/pull/40#issuecomment-1657120863) about having our own packages -- i just haven't gotten around to adding just yet

getchoo:
This might be revisited now that we vendor our own. It should be similar to `catppuccin.sources`

PR #591, opened 2025-06-28
feat(home-manager): add userstyle support for firefox-based browsers

different-name:
@isabelroses I spent the day today implementing [catppuccin-userstyles-nix](https://github.com/different-name/catppuccin-userstyles-nix) as a catppuccin/nix module

This adds userstyle (Stylus) support for Firefox based browsers supported by Home Manager's mkFirefoxModule helper (`firefox`, `librewolf` & `floorp`)

This falls under the option set

```
catppucin.<browser>.profiles.<profile>.userstyles
```

This allows for per profile configuration

Everything is under the `userstyles` option to allow for separate Firefox Color configuration options under the `theme` option (see #592)

Individual userstyles options can also be configured, for example:

```nix
catppuccin.firefox.profiles.default.userstyles = {
  # Configure individual userstyles
  settings = {
    youtube = {
      lightFlavor = "latte";
      darkFlavor = "mocha";
      accentColor = "mauve";

      logo = true;
      oled = false;
      sponsorBlock = true;
    };

    # Remove GitHub userstyle
    github.enable = false;
  };
};
```

Stylus must currently be installed by the user manually, but we have two other options for this: we could package and install Stylus ourselves, or do what [Stylix does with Firefox Color and use NUR](https://github.com/nix-community/stylix/blob/4ead8043f70cc3b951e704a1f6e40c8a10230e61/modules/firefox/each-config.nix#L104)

### DIFF
--- a/modules/home-manager/firefox.nix
+++ b/modules/home-manager/firefox.nix
@@ -6,6 +6,7 @@ let
     foldl'
     hasAttr
     importJSON
+    mapAttrs'
     toSentenceCase
     mapAttrs
     mkIf
@@ -18,6 +19,8 @@ let
 
   inherit (config.catppuccin) sources;
   themes = importJSON "${sources.firefox}/themes.json";
+
+  stylusExtensionId = "{7a7a4a92-a2a0-41d1-9fd7-1e92480d612d}";
 
   mkFirefoxModule =
     {
@@ -39,10 +42,11 @@ let
       cfg = getAttrFromPath modulePath config;
       firefoxCfg = getAttrFromPath hmModulePath config;
 
-      mkProfileOptions =
+      mkThemeOptions =
         {
-          forceDefault ? false,
-          enableDefault ? null,
+          name,
+          forceDefault,
+          enableDefault,
         }:
         catppuccinLib.mkCatppuccinOption (
           {
@@ -57,23 +61,101 @@ let
           force = mkOption {
             type = types.bool;
             default = forceDefault;
-            description = "Forcibly override any existing configuration for Firefox Color.";
+            description = "Forcibly override any existing configuration for ${name}.";
             example = true;
           };
+        };
+
+      mkBrowserOptions =
+        {
+          isProfile ? false,
+        }:
+        mkThemeOptions {
+          name = "Firefox Color";
+          forceDefault = if isProfile then cfg.force else false;
+          enableDefault = if isProfile then cfg.enable else null;
+        }
+        // {
+          userstyles =
+            mkThemeOptions {
+              name = "Stylus";
+              forceDefault = if isProfile then cfg.userstyles.force else false;
+              enableDefault = if isProfile then cfg.userstyles.enable else null;
+            }
+            // {
+              settings = mkOption {
+                type = types.attrsOf (
+                  types.submodule (
+                    { name, ... }:
+                    {
+                      options = {
+                        # we use mkOption instead of mkEnable option to default true
+                        enable = mkOption {
+                          type = types.bool;
+                          default = true;
+                          example = false;
+                          description = "Whether to enable ${name} Catppuccin userstyle";
+                        };
+
+                        # Since applying settings through home-manager makes the storage.js file read only,
+                        # build-stylus-settings removes the ability to configure userstyles through the stylus extension,
+                        # instead settings are defined here & are baked into the userstyles
+                        settings = mkOption {
+                          type = types.attrsOf (
+                            types.oneOf [
+                              types.str
+                              types.int
+                              types.bool
+                            ]
+                          );
+                          default = { };
+                          description = "Userstyle setting overrides";
+                          example = lib.literalExample ''
+                            {
+                              lightFlavor = "latte";
+                              darkFlavor = "mocha";
+                              accentColor = "mauve";
+                              logo = true;
+                              oled = false;
+                              sponsorBlock = true;
+                            };
+                          '';
+                        };
+
+                        # From what I can tell, this has no function with Catppuccin userstyles
+                        # It is probably better not to include this option, to reduce confusion
+                        # inclusions = mkOption {
+                        #   type = types.listOf types.str;
+                        #   default = [];
+                        #   description = "Custom included sites";
+                        #   example = "*://www.aaa.com/*";
+                        # };
+
+                        exclusions = mkOption {
+                          type = types.listOf types.str;
+                          default = [ ];
+                          description = "Custom excluded sites";
+                          example = "*://www.aaa.com/*";
+                        };
+                      };
+                    }
+                  )
+                );
+                default = { };
+                description = "Settings for Catppuccin userstyles";
+              };
+            };
         };
     in
     {
       options =
         (setAttrByPath modulePath (
-          (mkProfileOptions { })
+          (mkBrowserOptions { })
           // {
             profiles = mkOption {
               type = types.attrsOf (
                 types.submodule {
-                  options = mkProfileOptions {
-                    enableDefault = cfg.enable;
-                    forceDefault = cfg.force;
-                  };
+                  options = mkBrowserOptions { isProfile = true; };
                   config = {
                     flavor = lib.mkDefault cfg.flavor;
                     accent = lib.mkDefault cfg.accent;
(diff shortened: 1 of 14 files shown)


SchweGELBin:
> > Individual userstyles options can also be configured, for example:
> 
> Would it also be possible to add extra domains to userstyles? Like if I wanted to, say, add a particular instance of something like SearXNG.

I thought about it too and it would be awesome, but I didn't find a good solution to it.
Youtube for example has multiple section with domains, so we can't jyst append it after the first domain we find.

uncenter:
> > > Individual userstyles options can also be configured, for example:
> > 
> > 
> > Would it also be possible to add extra domains to userstyles? Like if I wanted to, say, add a particular instance of something like SearXNG.
> 
> I thought about it too and it would be awesome, but I didn't find a good solution to it. Youtube for example has multiple section with domains, so we can't jyst append it after the first domain we find.

Stylus allows you to specify more URLs to apply the userstyle to in a different section without editing the UserCSS. I wonder if you can use that and bake the custom URLs into the Stylus import file in the script or whatever?
CleanShot 2025-07-02 at 15 04 47

pkgs/build-stylus-settings/package.nix
```diff
@@ -0,0 +1,42 @@
...
+  };
+in
+buildNpmPackage (finalAttrs: {
+  pname = "build-stylus-settings";
+  version = "0.2.0";
+
+  src = builtins.path {
+    path = ./src;
+    name = finalAttrs.pname + "-src";
+  };
```
isabelroses, line 24: This is better imo https://noogle.dev/f/lib/cleanSource
```suggestion
  src = lib.cleanSource ./src;
```

pkgs/build-stylus-settings/package.nix
```diff
@@ -0,0 +1,42 @@
...
+    ./uncomment-stringify-type.patch
+  ];
+
+  npmDepsHash = "sha256-WqlsjB73awmqzuwQW33xt4W+XSojXQk+kLZ5EcV45+k=";
+
+  meta = {
+    description = "Stylus storage.js file generator, compatible with Catppuccin Userstyles";
+    license = lib.licenses.mit;
+    mainProgram = "build-stylus-settings";
+  };
```
isabelroses, line 41: you should add a maintaines field to prevent log spam for those using checkMeta

modules/home-manager/firefox.nix
```diff
@@ -4,17 +4,33 @@ let
...
+  stylusExtensionId = "{7a7a4a92-a2a0-41d1-9fd7-1e92480d612d}";
+
+  mkEnableOptionDesc =
+    description:
+    mkOption {
+      inherit description;
+      type = types.bool;
+      default = false;
+      example = true;
+    };
```
isabelroses, line 30: this is just `mkEnableOption` with a slightly different description. perhaps we can just change our description to  be better phrased for `mkEnableOption`.

see https://github.com/catppuccin/nix/pull/591/files#r2180810062

modules/home-manager/firefox.nix
```diff
@@ -42,12 +58,46 @@ let
                   accentSupport = true;
                 }
                 // {
-                  force = mkOption {
-                    type = types.bool;
-                    default = false;
-                    description = "Forcibly override any existing configuration for Firefox Color.";
-                    example = true;
-                  };
+                  force = mkEnableOptionDesc "Forcibly override any existing configuration for Firefox Color.";
```
isabelroses, line 61:
```suggestion
                  force = mkEnableOptionDesc "forcibly overriding any existing configuration for Firefox Color.";
```

So that would come out as

> Whether to enable forcibly overriding any existing configuration for Firefox Color.

modules/home-manager/firefox.nix
```diff
@@ -66,6 +116,37 @@ let
...
+                  # settings applied per userstyle as mentioned above
+                  # these take priority over global settings
+                  perStyleSettings = (
+                    listToAttrs (
+                      mapAttrsToList (name: value: {
+                        name = "Userstyle ${name} Catppuccin";
+                        value = mapAttrs (_: castToInt) value;
+                      }) profile.userstyles.settings
+                    )
+                  );
```
isabelroses, line 142: [mapAttrs'](https://noogle.dev/f/lib/mapAttrs') effectively does the same with less lines :D

```suggestion
                  perStyleSettings = lib.mapAttrs' (name: value: {
                    name = "Userstyle ${name} Catppuccin";
                    value = mapAttrs (_: castToInt) value;
                   }) profile.userstyles.settings;
```

modules/home-manager/firefox.nix
```diff
@@ -4,17 +4,33 @@ let
...
+  mkEnableOptionDesc =
+    description:
+    mkOption {
+      inherit description;
+      type = types.bool;
+      default = false;
+      example = true;
+    };
+
+  castToInt = v: if isBool v then (if v then 1 else 0) else v;
```
isabelroses, line 32: i think the lambda name here is a bit missleading.

modules/home-manager/firefox.nix
```diff
@@ -42,12 +58,46 @@ let
...
+                            darkFlavor = "mocha";
+                            accentColor = "mauve";
+
+                            logo = true;
+                            oled = false;
+                            sponsorBlock = true;
+                          };
+
+                          github.enable = false;
+                        };
```
isabelroses, line 98: this should really be put through [lib.literalExample](https://noogle.dev/f/lib/literalExample)

```suggestion
                        example = lib.literalExample ''
                          {
                          	youtube = {
                           	  lightFlavor = "latte";
                              darkFlavor = "mocha";
                              accentColor = "mauve";

                              logo = true;
                              oled = false;
                              sponsorBlock = true;
                            };

                            github.enable = false;
                          };
                        '';
```

pkgs/build-stylus-settings/package.nix
different-name: Should I just put you and getchoo? I am not on the nixpkgs maintainers list yet (but my first PR has one approval! so maybe I will be soon)

Happy to be listed, but I am not sure how that would work

different-name:
> Would it also be possible to add extra domains to userstyles? Like if I wanted to, say, add a particular instance of something like SearXNG.

> Stylus allows you to specify more URLs to apply the userstyle to in a different section without editing the UserCSS. I wonder if you can use that and bake the custom URLs into the Stylus import file in the script or whatever?

I will look into this in an hour or two, but maybe it's worth also adding an `extraConfig` option - this could be used to add more userstyles and define any stylus settings we dont end up defining here, which wouldnt otherwise be possible without using `sources` manually

modules/home-manager/firefox.nix
different-name: That's so cool :o I need to learn about more of lib's prime functions

different-name:
> Stylus allows you to specify more URLs to apply the userstyle to in a different section without editing the UserCSS. I wonder if you can use that and bake the custom URLs into the Stylus import file in the script or whatever?

Included sites is unfortunately pointless here, it doesn't have any effect on what URLs a given section will apply to and I believe all Catppuccin userstyles are limited to specific websites through the section rules, so it will never have an effect

Exuded sites is useful though, and we should include it, it's a simple property called `exclusions` at the root of the userstyle data, it's an array of strings

I don't see any other options we should include, the only settings users wouldn't be able to configure through extraOptions is auto update and light/dark mode, which don't do anything here

> Would it also be possible to add extra domains to userstyles? Like if I wanted to, say, add a particular instance of something like SearXNG.

Since the Catppuccin themes use rules, the normal way to do this through Stylus UI isn't through any settings, but actually to just modify the theme itself, which obviously doesn't work when you're dealing with updates (though I haven't looked at the how Stylus handles updating modified userstyles)

I think the best way to handle this is to just patch the catppuccin/userstyles source #186 seems related

The only other alternative I see is implementing our own system for patching just the rules, open to suggestions on that - we can't do it based on index, since that could break between updates, and so maybe text substitution? But I would just prefer using patch files at that point

different-name:
Pushing the requested changes, still need to implement `exclusions`, and I think I need to revisit the format of the JSON file that's passed to `build-stylus-storage`

I'm not too experienced with typescript, if anyone would be willing to review `build-stylus-storage` that would be amazing

different-name:
I'd like to keep things as generalized as possible, this seems like a better structure

```json5
{
  // override @var defaults
  "defaultSettings": {
    "lightFlavor": "mocha",
    "darkFlavor": "mocha",
    "accentColor": "mauve"
  },

  // per-userstyle settings
  "userstyles": {

    // Matched based on userstyle @name
    "GitHub Catppuccin": {

      // @var definitions
      "settings": {
        ...
      },

      // Custom included sites
      "inclusions": [
        ...
      ],

      // Custom excluded sites
      "exclusions": [
        ...
      ]
    },

    // Case insensitive
    "nixos wiki catppuccin": {
      
      // If false, the userstyle will be skipped
      "enable": false
    }
  }
}
```

uncenter:
> Included sites is unfortunately pointless here, it doesn't have any effect on what URLs a given section will apply to and I believe all Catppuccin userstyles are limited to specific websites through the section rules, so it will never have an effect

Ah gotcha. Wondering if there is anything I can change on the userstyles side of things to make this easier?

different-name:
I'm not familiar with userstyles, but it would be trivial if Stylus options could be used to define the rules

I'm not sure using any kind of variable (stylus or less) is possible though, since I think stylus parses sections before variables are applied? And apart from that I don't think anything could change upstream to make config simpler

Even if that did work, it's probably not a great idea to move where domain matching is configured

different-name:
I just realized, `extraConfig` isn't needed, because there's nothing stopping you from still defining options through the home-manager module

different-name:
What do we think of avoiding the hm module for the browser and using `home.file` directly to avoid IFD? IFD is pretty bad with this one, and it's kind of a waste to convert from JSON to Nix just to convert back to JSON again

different-name:
Closing this PR following feedback from a maintainer, who raised the following concerns:

- This reimplements Stylus’ import functions, which isn’t desirable
- Build helpers like this shouldn’t live in this repo; they should be upstreamed elsewhere
- They aren’t interested in reviewing the code or learning about how userstyles work

If support for all userstyles and preprocessors (not just Catppuccin and Less) were added, the build tool could have it's own repo & be packaged in nixpkgs. However, since this is essentially a reimplementation of Stylus’ import function, I don’t feel comfortable inflating the scope without using Stylus’ source directly

I’ve been working a lot with the Stylus source and experimenting with Webpack, but haven’t found a clean way to use the source without patches - other than emulating the based browser functionality in Node (which should be possible)

With that, and the maintainer’s stance on this PR combined with my hellish experience on the main Firefox module (and #676 reverting that work) leads me to believe the best chance of getting this merged would be to get as much implementation upstream as possible; create a new repo, rewrite the build tool to emulate a browser and use Stylus source, get it packaged in nixpkgs and get support for it in the Firefox Home Manager module. We would only need a tiny PR here then

I might tackle that at some point, but the scope is so large, and would would scrap everything here regardless, so I have decided to close this

Thank you Isabel for all your help on this and the Firefox module :pray: I've learnt an absolute ton of nix knowledge from it all

If anyone wants to try rewrite the build tool, feel free to ask any questions :heart: I'd be more than happy to help

Issue #580, opened 2025-06-04
Styling multiple profiles


MichaelXavier:
I'm using home-manager to manage vscode settings and have enabled catppuccin's vscode theme. One issue I'm seeing is that the home-manager module applies the theme settings to `programs.vscode.profiles.default`. If I have other profiles in that setting they don't get themed. VSCode doesn't seem to have any concept of profile inheritance so I'd ideally like to do that with nix attributes. I've tried defining another profile by merging with `config.programs.vscode.profiles.default` but that leaves to infinite recursion. Any suggestions would be greatly appreciated!

PR #595, opened 2025-07-03
feat(home-manager/vscode): Add support for per-profile config

different-name:
Closes: #580

Now that 25.05 is stable, I have removed configuration through home-manager's `programs.vscode` and switched to allowing per-profile config

Other than breaking support for home-manager versions below 25.05, there should be no breaking changes here:

- `catppuccin.vscode.*` options were renamed to `catppuccin.vscode.profiles.<profile>.*` using `lib.mkkRenamedOptionModule`
- `catppuccin.vscode.profiles` defaults to `{ default = { }; }`, maintaining auto enable behaviour


different-name:
:o no changes this time!

different-name:
Oh I didn't even realise lol, what's ctp?

isabelroses:
ctp is short for catppuccin

Issue #455, opened 2025-01-11
lazygit fails to open on first launch due to the assumption that config.yml exists


DrymarchonShaun:
Currently, the lazygit module assumes `config.yml` exists in its default location when it sets `$LG_CONFIG_FILE`, but that file doesn't get created until first launch, and if the aforementioned variable contains a file that doesn't exist, it fails to launch.
```
$ lazygit             
2025/01/10 20:31:53 stat /home/shaun/.config/lazygit/config.yml: no such file or directory
```

DrymarchonShaun:
I would have just submitted a PR, but I'm not entirely sure how to mitigate the issue.

natecox:
I was about to say that I couldn't replicate this, but then I realized that the config file is always being created for me via home-manager's install before a run can be tried. Out of curiosity, how are you installing lazygit?

uncenter:
> I was about to say that I couldn't replicate this, but then I realized that the config file is always being created for me via home-manager's install before a run can be tried.

I experienced this issue a few months ago (https://discord.com/channels/907385605422448742/1089733914328838294/1318055416684482590) because I deleted my (empty) lazygit config file for some forgotten reason, and home-manager actually *wasn't* recreating it...

getchoo:
> I suppose the ideal way would be to feed it into home-manager's programs.lazygit.settings but that would require parsing YAML to Nix, which doesn't seem to be officially supported yet?

We support this here via [importYAML](https://github.com/catppuccin/nix/blob/4a5ac694d7f8a63dec75cbe0ac1c84c818b6b789/modules/lib/default.nix#L81-L100)

It uses IFD which isn't great, but if there are no other alternative to applying the configuration in a mergeable (with other user settings) way, I'd say it's fine

DrymarchonShaun:
> It uses IFD which isn't great, but if there are no other alternative to applying the configuration in a mergeable (with other user settings) way, I'd say it's fine

Actually, looking at the commit history, it looks like this is how it was being done prior to #304.

getchoo:
Should be fixed by reverting that. Thanks!

PR #604, opened 2025-07-04
chore(lazygit): remove IFD

alejandro-angulo:
Prevents IFD without reintroducing the issue described in https://github.com/catppuccin/nix/issues/455

### DIFF
--- a/modules/home-manager/lazygit.nix
+++ b/modules/home-manager/lazygit.nix
@@ -1,9 +1,27 @@
 { catppuccinLib }:
-{ config, lib, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 let
   inherit (config.catppuccin) sources;
+
   cfg = config.catppuccin.lazygit;
+  enable = cfg.enable && config.programs.lazygit.enable;
+
+  # NOTE: On MacOS specifically, lazygit expects its configuration to be in
+  # `~/Library/Application Support` when not using XDG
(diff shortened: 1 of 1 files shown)


modules/home-manager/lazygit.nix
```diff
@@ -22,10 +40,16 @@ in
...
-    # TODO: Find a way to "source" a config file with $LG_CONFIG_FILE
-    # and *not* come across https://github.com/catppuccin/nix/issues/455
-    # https://github.com/jesseduffield/lazygit/blob/4b4d82e13c568f3b43c018740132454446cb0a2e/docs/Config.md#overriding-default-config-file-location
-    programs.lazygit.settings = catppuccinLib.importYAML (sources.lazygit + "/${cfg.flavor}/${cfg.accent}.yml");
+  config = lib.mkIf enable {
+    home.sessionVariables =
+      let
+        configFiles = [
+          "${sources.lazygit}/${cfg.flavor}/${cfg.accent}.yml"
+        ] ++ lib.optional (config.programs.lazygit.settings != { }) configFile;
```
alejandro-angulo, line 48: If I understand the issue in #455 correctly, the issue is that home manager doesn't write lazygit's config file if we have an empty attrset.

This will break if someone doesn't use home manager to manage their lazygit config file (and for some reason only manages the theme config with this flake ¯\\\_(ツ)\_/¯ ) but I think it's fine to assume users will manage the config with home manager.

Issue #518, opened 2025-03-24
Add support for vscode-icons


KP64:
Since we already have theming support for vscode (#509) it would also be nice to have an option to enable the catppuccin [vscode-icons](https://github.com/catppuccin/vscode-icons).
This would bring the whole catppuccin experience to vscode.

PR #619, opened 2025-07-23
feat(home-manager/vscode-icons): init

isabelroses:
closes https://github.com/catppuccin/nix/issues/518

### DIFF
--- a/modules/home-manager/vscode.nix
+++ b/modules/home-manager/vscode.nix
@@ -64,6 +64,8 @@ in
                   };
                 };
               };
+
+              icons = catppuccinLib.mkCatppuccinOption { name = "vscode-icons"; };
             };
         }
       )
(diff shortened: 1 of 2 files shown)


KP64:
Thank you for the work!
Some feedback, as the one that requested this.

1. I think that people, who are activating the icons, are expecting the icons to be activated immediately. We would need to set `workbench.iconTheme` just like we do with `workbench.colorTheme`.
2. What about flavors? I think one should be able to set them independently for the icons, like we are doing with e.g. `catppuccin.gtk.flavor`, where `catppuccin.flavor` is the default?

isabelroses:
> 1. I think that people, who are activating the icons, are expecting them to be activated immediately. We would need to set `workbench.iconTheme` just like we do with `workbench.colorTheme`.


thanks you this advice is exactly what i needed as a none-vscoder.

> 2. What about flavors? I think one should be able to set them independently for the icons, like we are doing with e.g. `catppuccin.gtk.flavor`, where `catppuccin.flavor` is the default?

as for this i left it out since theres no way i know of to make the icon theme declarative. this is mostly due to the code gen process of catppuccin/vscode-icons. and i spoke to praz about this and they said they would see what they could do.

KP64:
Works like a charm. Thank you!

PR #622, opened 2025-07-24
fix(foot): update port for v1.22.3

aciceri:
Upstream [has been udpated](https://github.com/catppuccin/foot/pull/20), without this change every time a foot terminal is started an error is printed.

@getchoo I'm assuming that `catppuccin` targets `nixos-unstable`, this change will now show an error when starting version `1.22.3` (on `nixos-25.05`)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -32,7 +32,25 @@ let
         directory = ./pkgs;
       };
     in
-    generated // collected;
+    generated
+    // collected
+    /*
+      TODO(@getchoo): Remove this with 25.05!!!
+
+      Pin foot to an older version in Nixpkgs < 25.11. See:
+      - https://github.com/catppuccin/nix/issues/636
+      - https://github.com/catppuccin/nix/pull/622
+    */
+    // lib.optionalAttrs (!(lib.versionAtLeast lib.version "25.11pre")) {
+      foot = generated.foot.overrideAttrs (oldAttrs: {
+        src = oldAttrs.src.override {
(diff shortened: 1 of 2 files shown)


isabelroses:
catppuccin nix supports both stable and unstable. so this is certainly a new unexpected issue

aciceri:
Are you aware of a similar situation for another package? I guess we could check the derivation's version and optionally use a different source.

But I don't know if putting two versions of the same source in the `sources.json` file is a good idea, probably the `paws.py` script wouldn't manage it well. Perhaps we could hardcode a `fetchFromGitHub` in the `<1.23` branch.

If you approve the approach I would be happy to implement it.

getchoo:
Honestly this is just the straw that broke the camel's back

We could probably do a silly hack where we check `lib.version` and choose the older source for stable, but that's really a band-aid fix on top of a ton of others. We're going to be splitting the repo into unstable and stable branches, just to get rid of these types of compat issues

Work for this (and a few other things) is now ongoing in https://github.com/catppuccin/nix/pull/627. I hope to have it done by Tuesday (07/29) as to now block this any further, or anything else that might come up

getchoo (approved):
To keep compatibility with 25.05 for a little bit longer on `main`, I've applied the above hack to pin the source with Nixpkgs < 25.11pre. This will be removed along with 25.05 support on `main` in the near future, and hopefully PRs like this can just be merged immediately now without all the fuss

Thanks again!

PR #627, opened 2025-07-26
docs: move to starlight

getchoo:
~~This is work going to be done alongside what's going on in the `site` branch~~

~~The idea here is to fix our backwards compatibility issue by simply emulating NixOS/Nixpkgs: main will only support `nixos-unstable`, and we will have release branches corresponding to each NixOS/home-manager release.~~

~~As this gives us the opportunity to seriously clean up the complication in things like our development Flake and CI, I also want to take this opportunity to get our long overdue Starlight port going. The aforementioned `site` branch will handle this with a more traditional deployment style than our current one - mainly because it would be a pain to constantly bump the hashes of each branch in a Flake for example. This will also replace our current option search, by eating up some JSON we now spit out in the development Flakes of each branch~~

This is just porting our docs to [Starlight](https://starlight.astro.build) and making some preparations for the eventual stable branch-off


CHANGELOG.md
isabelroses, line 1: I see this file got formatted, maybe we could add it to the excludes list since its a generate file?

isabelroses (approved):
LGTM, one nit really :D

LovingMelody:
This change broke my nix-config I'm assuming from commit 8b9f0bc  

```
The option `home-manager.users.melody.programs.firefox.profiles' in `/nix/store/r833dgknjqyv4y6snhqgn5cy0mvcjp6n-source/modules/programs/firefox' is already declared in `/nix/store/d9hd7jqywksl91rsaqah9phc4vskcrpi-source/modules/home-manager/firefox.nix'.
```
`r833dgknjqyv4y6snhqgn5cy0mvcjp6n-source` is home-manager
`d9hd7jqywksl91rsaqah9phc4vskcrpi-source` is catppuccin/nix

PR #629, opened 2025-07-29
fix: useFetchCargoVendor is non-optional and enabled by default as of 25.05

Safenein:
This pull request makes a small change by removing the `useFetchCargoVendor` attribute from two `rustPlatform.buildRustPackage` definitions.

It was raising a warning like this one:

```
evaluation warning: buildRustPackage: `useFetchCargoVendor` is non‐optional and enabled by default as of 25.05, remove it
```

So here this fix follows the given instruction.

### Package definition updates:
* Removed the `useFetchCargoVendor` attribute from the `rustPlatform.buildRustPackage` definition in `pkgs/catwalk/package.nix`.
* Removed the `useFetchCargoVendor` attribute from the `rustPlatform.buildRustPackage` definition in `pkgs/whiskers/package.nix`.

### DIFF
--- a/pkgs/catwalk/package.nix
+++ b/pkgs/catwalk/package.nix
@@ -18,7 +18,6 @@ rustPlatform.buildRustPackage rec {
     hash = "sha256-Yj9xTQJ0eu3Ymi2R9fgYwBJO0V+4bN4MOxXCJGQ8NjU=";
   };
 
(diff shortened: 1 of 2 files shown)


Safenein:
You are welcome. 
Please, have a Nice day.

Issue #602, opened 2025-07-04
Firefox module: default is still unconditional, please do not dismiss underlying issue!


kittywitch:
Actual behaviour not addressed:
* default is unconditionally added to profiles, regardless of per-profile configuration!

Without global enable:
       Failed assertions:
       - kat profile: Must have exactly one default Firefox profile but found 2, namely default, main
       - kat profile: Must not have a Firefox profile with an existing ID but
  
Regards forcibly closed https://github.com/catppuccin/nix/issues/601

@isabelroses: The actual per-profile toggle does not disable the underlying behaviour of enabling a profile regardless of what occurs; profiles.default will be created regardless, this means that simply adding the firefox module into your home manager modules changes the underlying profile setup!

isabelroses:
> It redefines default to provide profiles.default = {}; as existent UNCONDITIONALLY.
> I think the only way around this would be to make sure
> catppuccin.firefox.profiles = {};

But this was my point in your last issue. And on that note don't just make new issues if they are closed. Make your point in that issue and I would have reopened it.

kittywitch:
My point is that I do not even enable it for firefox (when having the global enable for catppuccin set to false), and yet it pollutes programs.firefox.profiles.default regardless?

This point was supposed to have already been made in the prior issue with the assertion raised from home-manager's firefox module. I am sorry for handling this by opening a new issue, but I saw you as dismissive of actual underlying problem that is caused for those of us who have a firefox profile already set as our default.

Firefox profiles in home-manager cannot be disabled without mkForce.

kittywitch:
like, the problem is, even if I had used catppuccin/nix in a completely fresh state, without any manual enables; this would still define `programs.firefox.profiles.default`, and that is bad!

brckd:
The same issue was [solved in Stylix](https://github.com/nix-community/stylix/pull/191) by adding a `profileNames` option. There is a warning in place to assure this option is properly configured. Daniel also [suggested](https://github.com/nix-community/stylix/pull/191#discussion_r1426700476) to add an option for shared configurations to Home Manager, similar to `home-manger.sharedModules`. I can create a PR for Home Manager if desired!

isabelroses:
I can create a PR for Home Manager if desired!

Thanks for the offer, but #606 should have fixed this. I have just left the issue open in case the fix did not truly work. but i think it should have.

different-name:
[...]

Issue #601, opened 2025-07-04
Firefox module: make default profile behaviour able to be disabled, or not default, or anything but unconditional?


kittywitch:
```
       - kat profile: Must have exactly one default Firefox profile but found 2, namely default, main
```
Behaviour introduced with https://github.com/catppuccin/nix/commit/423c448b6823cb6d0c284116d3829b16120ef0d9

Today I was greeted with this after a `nix flake update`. Please do *not* just straight up enable the default profile for anyone who uses the module unconditionally without providing a way of opting out of this behaviour? Some of us do not particularly want the attr "default" forced upon us or to have to migrate our Firefox profile as a result of a theme change, or to have to mkForce all of our Firefox home-manager config.

isabelroses:
> Please do not just straight up enable the default profile for anyone who uses the module unconditionally without providing a way of opting out of this behaviour?

This is not true though. Through `catppuccin.firefox.main.enable = false` to disable it. Furthermore you opted into the "global enable" to have triggered this error in the first place.

kittywitch:
You are outright incorrect; issue reopened as https://github.com/catppuccin/nix/issues/602

The actual issue is that profiles.default is created regardless of any enablement of the module. The profile addition is not dependent upon *anything* but the import of home-manager modules from this flake. I tested this with the global enable set to false.

PR #633, opened 2025-07-30
fix(home-manager/firefox): only apply to profiles specified

### LINKED ISSUE #631: add a way to disable firefox modules
i can't find a way to disable the floorp module (though i'm sure this applies to firefox and librewolf) which is a problem because it conflicts with the rest of my config

getchoo:
~~Generally, using conditional option merging to apply these settings at
the submodule level:~~

~~- Breaks standalone evaluation of the homeModule (like in our docs)~~
~~- Allows for strange effects that only happen conditionally, like https://github.com/catppuccin/nix/pull/627#issuecomment-3134735931~~
  
The above is still a point I'd like to make, but conditional option merging is required here. We just need to be careful with it

Fixes #631, as well as a regression from #627

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -33,6 +33,10 @@ jobs:
         run: |
           nix fmt
 
+      - name: Install Node modules
+        run: |
+          nix develop --command pnpm install
+
       - name: Format with Prettier
         run: |
           nix develop --command nrr fmt
--- a/dev/flake.nix
+++ b/dev/flake.nix
@@ -59,7 +59,8 @@
           in
 
           (pkgs.nixosOptionsDoc {
-            options = lib.removeAttrs eval.options [ "_module" ];
+            # You can also use this for backwards compat with < v1.2: `options = lib.removeAttrs eval.options [ "_module" ];`
+            options = { inherit (eval.options) catppuccin; };
 
             transformOptions =
               opt: lib.recursiveUpdate opt { declarations = replaceDeclarations opt.declarations; };
--- a/modules/home-manager/firefox.nix
+++ b/modules/home-manager/firefox.nix
@@ -101,11 +101,6 @@ let
                 }
               )
             );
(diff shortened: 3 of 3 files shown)


different-name:
@getchoo this seems to essentially revert #609 which was the solution to #602

Also see #601 and #606

I believe the implementation in this PR would render global enable non functional?

different-name:
> And this specific implementation also has unintended side effects, like
applying the extension settings to profiles not actually specified in
the catppuccin.{firefox,floorp,librewolf}.profiles option

This is the intended feature of #609, these profiles are not touched without global enable being enabled

If global enable isn't used, profiles must be specified manually

`catppuccin.firefox.*` can be used to control default configuration for all profiles, so `catppuccin.firefox.enable = false;` will default disable the theme for all profiles, even if global enable is used

getchoo:
> I believe the implementation in this PR would render global enable non functional?

Yep, totally missed the previous work here. My bad!

> If global enable isn't used, profiles must be specified manually

> catppuccin.firefox.* can be used to control default configuration for all profiles, so catppuccin.firefox.enable = false; will default disable the theme for all profiles, even if global enable is used

I think I've figured out a good middle ground here: by default `catppuccin.firefox.profiles` will be populated by all of those in `programs.firefox.profiles`, and as it is currently, these profiles will respect the global `catppuccin.firefox.enable` unless otherwise specified. However, a guard will now be put into place to ***only apply changes to profiles defined in our `catppuccin.firefox.profiles` option***. With the aforementioned default for `catppuccin.firefox.profiles` this shouldn't change the practical OOTB experience with global enable applying to all profiles by default, ***unless `catppuccin.firefox.profiles == { }`***

There will also be a warning for any profiles defined in `catppuccin.firefox.profiles` that do not have an accompanying entry in `programs.firefox.profiles`, since this doesn't have any effect and isn't immediately obvious (like in our CI!)

PR #632, opened 2025-07-30
fix(tests): update firefox module test options for #609

different-name:
Since #609 the test options no longer work for testing the Firefox module, hence the missed build failure in https://github.com/catppuccin/nix/pull/627

~~This PR should result in a build failure currently, since the Firefox module is broken by 8b9f0bc231c854c78eb719921ca6831ed28003b9~~ fixed in https://github.com/catppuccin/nix/commit/2f0b779c531ac922a3f0e3f817ff4e794865873a

PR #634, opened 2025-07-31
ci: evaluate instead of building configurations, dogfood configurations for doc gen

getchoo:
Recently I've come to the conclusion that our testing setup is a *lot* of effort and resource use relative to what we actually get out of it

The main issues I see are:

- It can't really take into account many variants
  - Given we have one NixOS configuration and (mostly) one home-manager configuration, our testing coverage can't really include cases where one option is enabled *and* disabled for example
  - This leads to issues like https://github.com/catppuccin/nix/pull/627#issuecomment-3134735931, where a specific conditional we aren't currently testing with presents serious problems
- It takes a ton of time
  - Nearly all of our supported programs need to be downloaded onto the test machine at once. This usually takes up more time than actually booting up and running the VM
  - Rebuilds are triggered for *any* change to *any* port. The time cost mentioned above now happens on nearly every PR
  - Building a full GUI system leaves us much more vulnerable to the all-too-common Nixpkgs build failures, especially on Darwin. This leads to even more egregious CI times, which usually fail since that's why they're not cached in the first place
  - See https://github.com/catppuccin/nix/pull/608 for some cool numbers around this (and thanks @kitsiosk for inspiring these changes a bit!)
- It doesn't actually catch many basic failures
  - Since we aren't running any of the applications, we aren't coming across the errors our users are
  - Even if we *were* to try to run some applications, it's still difficult to tell the colorscheme has even applied screenshotting each one and (I'd imagine in many cases) manually verifying
  - https://github.com/catppuccin/nix/pull/622 is an example of this, where our CI didn't catch the error immediately printed on launching `foot`
  - This might be cool to do for specific, complicated modules some day though. Having single module tests would also make all the other points here basically moot!
- It's annoying to maintain
  - Enabling every single module we support sucks. I don't think anyone here likes it - or at least I know I don't :/
  - This is also forgotten about sometimes, see https://github.com/catppuccin/nix/pull/632

The solution I'm implementing here is to only **evaluate** these configurations instead. I believe evaluation is the only practical advantage we've had in our current CI setup for a while, and after making it a *bit* more strict by running everything through `nixosOptionsDoc`, I think it's better than ever. Hopefully this (and some better handling of things like option merging that can't be easily worked around on the user's end) will help prevent some of the recent mistakes, and keep our focus more on what we can prevent currently

Our testing configurations are also nicely exported in the dev flake under `homeConfigurations.pepperjack` and `nixosConfigurations.pepperjack-pc` now, so it should be easier to play around with and debug the modules in a repl.


isabelroses (approved):
LGTM

PR #642, opened 2025-08-01
chore: deprecate 25.05

getchoo:
Not to be backported, obviously

### DIFF
--- a/modules/global.nix
+++ b/modules/global.nix
@@ -8,13 +8,14 @@
 
 let
   catppuccinLib = import ./lib { inherit config lib pkgs; };
+
+  minimumVersion = "25.11";
+  isMinimumVersion = lib.versionAtLeast catppuccinLib.getModuleRelease minimumVersion;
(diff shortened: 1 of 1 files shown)


modules/global.nix
```diff
@@ -8,13 +8,14 @@
 
 let
   catppuccinLib = import ./lib { inherit config lib pkgs; };
+
+  minimumVersion = "25.11";
+  isMinimumVersion = lib.versionAtLeast catppuccinLib.getModuleRelease minimumVersion;
+  # Use this to toggle between a deprecation warning for a release and a full assertion
+  warn = true;
```
isabelroses, line 15: imo this should be an option. particularly when comparing to [nix-darwin](https://github.com/nix-darwin/nix-darwin/blob/e04a388232d9a6ba56967ce5b53a8a6f713cdfcf/eval-config.nix#L22) and [home-manager](https://github.com/nix-community/home-manager/blob/1d7abbd5454db97e0af51416f4960b3fb64a4773/modules/home-environment.nix#L536)

PR #670, opened 2025-08-04
refactor(home-manager/firefox): factor out options into submodules

getchoo:
Along with separating our `mkFirefoxModule` function, this should help
cleanup the module as whole, while also making it much more
understandable

CC @different-name


modules/home-manager/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,134 @@
+{
+  catppuccinLib,
+
+  name,
+  hmModulePath ? [
+    "programs"
+    name
+  ],
+  ...
```
different-name, line 9: Why do we accept additional arguments here?

modules/home-manager/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,134 @@
...
+  themes = importJSON "${sources.firefox}/themes.json";
+
+  cfg = getAttrFromPath modulePath config;
+  firefoxCfg = getAttrFromPath hmModulePath config;
+
+  modulePath = [
+    "catppuccin"
+    name
+  ];
+  prettyName = if args.prettyName or null == null then toSentenceCase name else args.prettyName;
```
different-name, line 42: I feel like we should use

```nix
prettyName ? null
```

In the function args instead, being strict with the typing of the arguments is better for self documenting & clarity

modules/home-manager/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,134 @@
...
+          type = types.bool;
+          default = false;
+          description = "Forcibly override any existing configuration for Firefox Color.";
+          example = true;
+        };
+      };
+  };
+
+  catppuccinProfilesSubmodule = {
+    imports = [ optionsModule ];
```
different-name, line 61: nit: styling is different to the imports statement for `catppuccinOptions` below

```nix
imports = [
  optionsModule
];
```

modules/home-manager/firefox/default.nix
```diff
@@ -0,0 +1,25 @@
+{ catppuccinLib }:
+{ lib, ... }:
+
+let
+  mkFirefoxModule =
+    args:
+
+    lib.modules.importApply ./mkFirefoxModule.nix (
```
different-name, line 8: Is the reason for separating this out into a new file so that it could be imported & used elsewhere? (even if it's external)

different-name (changes requested):
Otherwise LGTM, haven't had the chance to test it yet

different-name (approved):
haven't tested this, but looks good to me! thank you again getchoo!

PR #693, opened 2025-08-15
fix(home-manager/freetube): fix theme setting formatting

Anomalocaridid:
The behavior of `lib.toSentenceCase` seems to have changed, which broke how the settings for `mainColor` and `secColor` were formatted. (e.g. What should have been CatppuccinMochaMauve became Catppuccinmochamauve)

This slightly changes how FreeTube support is implemented to account for this.


isabelroses (approved):

isabelroses:
thanks for the fix

PR #710, opened 2025-08-27
docs: Update obsolete flake attribute for importing catppuccin

MonsterDruide1:
`trace: evaluation warning: Obsolete Flake attribute 'catppuccin.homeManagerModules.catppuccin' is used. It was renamed to 'catppuccin.homeModules.catppuccin'.`

The first two examples on the page have it correct, but in the "end flake", it's using the old attribute.


isabelroses (approved):
Thanks :D

PR #711, opened 2025-08-27
fix(vscode): flakey builds

isabelroses:
adding the `--dev` flag makes the build reproducable. but we lose the minification going from a size of `124.46 KB` to `131.42 KB` which is an acceptible cost for not having to trigger CI 3 times to get a good build.


sgoudham:
I'd probably just wait until I make a stable release including https://github.com/catppuccin/vscode/pull/578 so that you won't have to change this again.

PR #735, opened 2025-09-26
feat(nixos): add support for cursors and icons

musjj:
Adds support for cursors and icons settings in the NixOS modules.

The general use-case here is for applying the cursor and icons settings to the login manager. I've only added support for GNOME because that's what I use, but support for other login managers could be added in the future.

### DIFF
--- a/modules/nixos/gtk.nix
+++ b/modules/nixos/gtk.nix
@@ -1,0 +1,49 @@
+{ catppuccinLib }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.catppuccin.gtk;
+in
+{
+  options.catppuccin.gtk = {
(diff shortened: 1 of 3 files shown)


modules/nixos/gtk.nix
```diff
@@ -0,0 +1,49 @@
...
+        cfg.icon.enable
+        && (config.services.desktopManager.gnome.enable || config.services.displayManager.gdm.enable)
+      )
+      {
+        services.displayManager.environment.XDG_DATA_DIRS = (
+          (lib.makeSearchPath "share" [
+            (pkgs.catppuccin-papirus-folders.override { inherit (cfg.icon) accent flavor; })
+          ])
+          + ":"
+        );
```
isabelroses, line 33: is there a reason we end with `:`

isabelroses:
oops. i really thought i merged this. my bad 😅

Issue #392, opened 2024-11-30
Tracking: removal of IFD in modules


getchoo:
[Import from derivation](https://nix.dev/manual/nix/2.25/language/import-from-derivation.html) is a cool and very helpful feature of Nix, but can seriously hurt evaluation performance, and may not be allowed in some cases. We should try to avoid it when possible (in a reasonable way)

Modules that currently require IFD include:

- [x] cava https://github.com/catppuccin/nix/issues/376, https://github.com/catppuccin/nix/pull/750
- [ ] firefox
- [ ] fzf
- [ ] gh-dash
- [ ] imv https://github.com/catppuccin/nix/issues/375
- [x] lazygit https://github.com/catppuccin/nix/pull/604
- [x] mako https://github.com/catppuccin/nix/issues/341, https://github.com/catppuccin/nix/pull/684
- [x] sioyek https://github.com/catppuccin/nix/pull/686
- [ ] starship
- [ ] skim
- [ ] swaylock https://github.com/catppuccin/nix/issues/340
- [ ] tty
- [x] wezterm
- [ ] gemini-cli

mightyiam:
Would it still be IFD if sources are flake inputs?

getchoo:
No, but for the same reason https://github.com/catppuccin/nix/pull/377 prevents IFD. Flake inputs use builtin fetchers, which would still block evaluation for every port download

See also https://github.com/catppuccin/nix/issues/388

stasjok:
I've got 'allow-import-from-derivation' error after updating from here:

https://github.com/catppuccin/nix/blob/8b6baed3d010416da6295805012b12cfa05c09f7/modules/home-manager/fzf.nix#L8

Searching 'importJSON' I found similar usages. Those modules can be added here I think.

multivac61:
Getting errors in CI after recent update.

getchoo:
Thanks for catching these, The list has been updated 👍

isabelroses:
As a note I would recommend just disabling starship if you have issues with IFD since starship should match your terminal colours.

getchoo:
IFD was re-introduced in the lazygit module with https://github.com/catppuccin/nix/pull/515

mightyiam:
In case anyone cares about `nix flake show` as much as I do, [stylix](https://stylix.danth.me/) does not inherently involve IFD.

[...]

PR #738, opened 2025-09-28
Expose color palette as attributes

### LINKED ISSUE #734: Expose color pallete as attributes
Sometimes you might want to quickly splash a little bit of color in your config, for example for a small plugin in VS Code. 

By exposing the color pallete as an attribute the user can easily do this without making a whole new port.

musjj:
Expose a set of read-only attributes containing the chosen Catppuccin color palette.

Example usage:

```nix
{ config, ... }:
{
  my-color = config.catppuccin.palette.colors.base.hex;
}
```

This will help eliminate IFD for modules that require access to the color palette for various reasons. Users who wants to add personal touches/customizations to their themes will also benefit from this.

The workflow for the palette generation is meant to be run manually (via `workflow_dispatch`), because I don't think it needs frequent/regular updates. But I can change that if necessary.

I'm not confident with how the files are organized, but I've decided to create a separate file for the palette types because it got a bit long. Also, there might be a way to automatically generate the Nix types from the TypeScript types, but it's written manually by hand for now.

Fix: https://github.com/catppuccin/nix/issues/734
Ref: https://github.com/catppuccin/nix/issues/392


modules/global.nix
```diff
@@ -46,6 +46,12 @@ in
         description = "Port sources used across all options";
       };
 
+    palette = lib.mkOption {
+      type = (import ./palette/type.nix) lib;
```
isabelroses:
```suggestion
      type = import ./palette/type.nix lib;
```

.github/workflows/palette.yml
isabelroses, line 1: lets remove this CI and add a special case to `pkgs/paws.py` that downloads the pallete and we just import the pallete.json into the module you made before.

isabelroses:
sorry for the long time to review. wasn't sure if i wanted this. decided it might not be the *worst* idea. still not exactly a fan but hey

isabelroses:
this addresses IFD in:
- tty
- fzf
- skim

is it worth all that extra work?

isabelroses:
actually i feel a lot more confident on the latest changes having seen catppuccin/rust

budimanjojo:
Would really like to get this merged. @musjj Can you resolve the file conflicts? I'm willing to do it if you aren't interested anymore.

isabelroses:
we (me and getchoo) are a bit concerned about the perf implications of this change. so even if the conflicts are resolved it may not move much. we were talking about perhaps not exposing it via modules and rather via lib

budimanjojo:
@isabelroses Why would it affect performance?

isabelroses:
we would be loading a file that is 3,635 lines long into your config. it will inevitably hurt evaluation speed no matter what. one way we could reduce that a bit is without using the module system, i.e. using lib instead but even that is going to only improve performance slightly

budimanjojo:
Isn't it what we're currently doing? The difference is this file is now available locally instead of from another repo?

I completely agree that a better approach to improve performance is needed but getting this merged will fixed the IFD issue for 3 modules and the performance improvement can be done in a different commit. I believe you and the other maintainers know better what's the right approach, but I hope IFD issue can be prioritized.

isabelroses:
from my understanding of nix we are currently lazily loading the json so people who don't use those ports will not be affected. but if that changes they will. fortunately as well nix has import caching so despite us loading the same file 3 times it will only appear on a eval profiler once

budimanjojo:
Ah I see, I think this PR can be changed to do it the current way too without much work.
The user should not be able to define their own color palette (which this PR provides).

budimanjojo:
@isabelroses What about sanitizing the json file inside the `paws.py` to process the file with `jq` to make the file smaller?

I see that those 3 modules (fzf, skim, tty) only needs the `hex` value. Something like this `jq` command:
```bash
jq '{
  mocha: .mocha,
  frappe: .frappe,
  latte: .latte,
  macchiato: .macchiato
}
| with_entries(
    .value = {
      colors: (
        .value.colors
        | with_entries(
            .value = { hex: .value.hex }
          )
      )
    }
  )' palette.json
```

Will result in a file with 330 lines like this:
```json
{
  "mocha": {
    "colors": {
      "rosewater": {
        "hex": "#f5e0dc"
      },
      "flamingo": {
        "hex": "#f2cdcd"
      },
      ...
    },
  },
}
```
Doing the sanitizing using nix after fromJSON will be easier, but if performance is wanted then sanitizing the source json file will be so much more efficient.

stefanboca:
`jq` also has a compact output mode, enabled by the `-c` flag, which removes whitespace. IME using compact mode cuts the resulting file size in ~half.

budimanjojo:
I don't think it's about the file size, but more about converting the JSON code into Nix code?

isabelroses:
> @isabelroses What about sanitizing the json file inside the `paws.py` to process the file with `jq` to make the file smaller?
> 
> I see that those 3 modules (fzf, skim, tty) only needs the `hex` value. Something like this `jq` command:

I assume the purpose of this pr was also to expose the config to users. so they might need more than just the hex values.

> I don't think it's about the file size, but more about converting the JSON code into Nix code?

`fromJSON` is relatively fast thats not really been the issue. the issue is the module system.

Issue #430, opened 2024-12-29
Is qt.platformTheme.name = "kvantum" really needed?


WildPowerHammer:
After some hours of trying to get icons in qt apps, I understood that qt5/6ct is needed or other platform theme plugin, it sets using QT_QPA_PLATFORMTHEME, which sets to kvantum by qt.plaformTheme.name = "kvantum" which is required [here](https://github.com/catppuccin/nix/blob/8b6baed3d010416da6295805012b12cfa05c09f7/modules/home-manager/kvantum.nix#L68) 

If i get [this comment](https://github.com/tsujan/Kvantum/issues/221#issuecomment-401080094) right, QT_QPA_PLATFORMTHEME=kvantum don't do anything useful, not sure, but i think it didnt changed since 2018 and as far as i can see unsetting this dont break anything because QT_STYLE_OVERRIDE still working

So, maybe it's worth to remove requirement for qt.platformTheme.name in catppuccin.kvantum?

getchoo:
> So, maybe it's worth to remove requirement for qt.platformTheme.name in catppuccin.kvantum?

Maybe?

If I'm being totally honest, I'm not personally familiar with either Qt theming or Kvantum. We've had a lot of issues with it in the past in regards to even getting things to apply

I believe this was brought up before in https://github.com/catppuccin/nix/issues/275 ...but the proposed solution in this issue (if I'm understanding it correctly) to only set QT_STYLE_OVERRIDE also [has issues](https://github.com/catppuccin/nix/issues/288)

If you or anyone else reading this knows more about this, feel free to chime in and/or open a PR. It'd be super appreciated :)

theNoof:
hello, same issue here. but when i set it, i get following error:
```nix
error: The option `home-manager.users.noof.qt.style.name` is defined both null and not null, in `/nix/store/m5nxipkc38w9g9wz23skgql43dxsblnz-source/modules/qt/hm.nix' and `/nix/store/6w5f2jqg8pr4l2wg6mw57058wpzp8xp3-source/programs/wm/catppuccin'.
```

musjj:
The assertion is not only unnecessary but just plain wrong, at least for non-KDE systems like mine. After hours of trial and error, I've figured out the correct incantation:

```nix
qt = {
  enable = true;
  style.name = "kvantum";
  platformTheme = {
    name = "qt5ct:qt6ct";
    package = with pkgs; [
      libsForQt5.qt5ct
      kdePackages.qt6ct
    ];
[...]

PR #741, opened 2025-10-03
fix(home-manager/kvantum): remove platform theme assertion

musjj:
As discussed here, setting `QT_QPA_PLATFORMTHEME` may be unnecessary: https://github.com/catppuccin/nix/issues/430#issuecomment-3346642996

But more importantly, the option enforced by the assertion simply does not work for certain platforms. This will leave the user's system in a broken state while at the same time preventing that very user from fixing it. It's *very* frustrating situation to be in.

For now, I propose simply removing the assertion until we have more data on what works and what doesn't.

On the other hand, I chose to leave the `QT_STYLE_OVERRIDE` assertion untouched because I believe it should be correct based on my experience and various online discussions I've come across.

If anyone have experience with setting Kvantum across different platforms (e.g. GNOME, KDE, plain `wlroots` compositors, etc.), I'd appreciate your feedback here.

### DIFF
--- a/modules/home-manager/kvantum.nix
+++ b/modules/home-manager/kvantum.nix
@@ -64,10 +64,6 @@ in
         ];
         message = ''`qt.style.name` must be `"kvantum"` to use `qt.style.catppuccin`'';
       }
-      {
-        assertion = lib.elem (config.qt.platformTheme.name or null) [ "kvantum" ];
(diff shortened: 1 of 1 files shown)


isabelroses (approved):
wow. thanks for the incredible work on this!

PR #755, opened 2025-10-20
fix(home-manager/delta): use new options

alapshin:
Update catppuccin-nix delta's config for https://github.com/nix-community/home-manager/pull/8006

### DIFF
--- a/modules/home-manager/delta.nix
+++ b/modules/home-manager/delta.nix
@@ -5,7 +5,7 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.delta;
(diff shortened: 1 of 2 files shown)


modules/home-manager/delta.nix
```diff
@@ -22,9 +23,9 @@ in
   };
 
   config = lib.mkIf enable {
-    programs.git = {
+    programs.git = lib.mkIf enableGitIntegration {
```
isabelroses:
```suggestion
    programs.git = lib.mkIf config.programs.delta.enableGitIntegration {
```

PR #774, opened 2025-11-27
feat(home-manager/cursors): add options to build specific cursor themes

KneeCapStealer:
These option allow for manually selecting which flavors and accents to build, decreasing build times for users.

I made it so the lists of flavors and accents is empty, and the code in the `buildPhase` then checks if they are empty and simply outputs the default `just all`, if that is the case.

The code can be quite smaller if the lists are filled out by default, although this makes the beginning of the file quite a bit longer, as the formatter puts each element on a new line.

It would be nice to have the `flavors` and `accents` parameters set by default via `catppuccin.cursors`, as the build times are quite long. But I am unsure if this would break things, so I haven't done that yet.

### DIFF
--- a/pkgs/cursors/package.nix
+++ b/pkgs/cursors/package.nix
@@ -9,6 +9,8 @@
   xcur2png,
   xorg,
   zip,
+  flavors ? [ ],
+  accents ? [ ],
 }:
 
 buildCatppuccinPort (finalAttrs: {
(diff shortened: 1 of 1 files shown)


HeroGamers:
I have the cachix added, but for some reason, I almost always end up having to spend 12-15 minutes waiting for all the cursors to build. Don't know why tbh. - guessing (without knowing) because I've set the flake to follow my own nixpkgs (unstable)?

PR #800, opened 2026-01-06
fix(vicinae): updated vicinae settings to new format

xaiyadev:
Vicinaes settings won't load because they changed their formatting about themes
this PR fixes this problem and uses the new updated format


isabelroses:
We really need to cut that 25.11 release. :sob:

isabelroses (approved):
LGTM. ty.

PR #817, opened 2026-01-07
feat(global): add `autoEnable` and make `enable = false` disable all modules

### LINKED ISSUE #796: `catppuccin.enable = false` doesn't disable everything
Hi,

First thanks for your work on this module, I've never been as aesthetically soothed when using my PC!

I think the global `enable` option is a bit confusing/unconventional: usually it disables all the config driven by options in the same attrset, but here it's just a default state for the other modules.

Example:
```nix
catppuccin.enable = false;
catppuccin.whatever.enable = true;
```

I'd expect this to not have any effect due to `enable = false`, as with essentially all modules out there, but currently the `whatever` program still has configuration installed.

I think basically the current behavior of `enable` should be moved to something like `catppuccin.autoEnable`, so that `enable` can be an actual kill switch. `autoEnable` is the name `stylix` uses, and it seems to me convenient to reuse it.  
I'm willing to make a PR for this if you're interested, but just want to check if you're interested before spending time on it!

ThinkChaos:
Fixes #796

### DIFF
--- a/modules/global.nix
+++ b/modules/global.nix
@@ -17,7 +17,20 @@ in
   imports = catppuccinLib.applyToModules catppuccinModules;
 
   options.catppuccin = {
-    enable = lib.mkEnableOption "Catppuccin globally";
+    enable = lib.mkEnableOption ''
+      Catppuccin.
+
+      Note: for `stateVersion` < 26.05, this is equivalent to `autoEnable`,
+      and there is no way to globally disable Catppuccin.
+      Since 26.05, this is a more traditional `enable` option;
+      `false` disables all Catppuccin modules.
+    '';
+
+    autoEnable = lib.mkEnableOption ''
+      all Catppuccin integrations by default.
+
+      Note: for `stateVersion` < 26.05, this option is equivalent to `enable`.
+    '';
 
     flavor = lib.mkOption {
       type = catppuccinLib.types.flavor;
@@ -56,6 +69,13 @@ in
         the version does not match catppuccin/nix's minimum supported version.
       '';
     };
+
+    # TODO: remove this back-compatibility implementation detail
+    _enable = lib.mkOption {
+      type = lib.types.bool;
+      internal = true;
+      readOnly = true;
+    };
   };
 
   config = {
@@ -70,7 +90,23 @@ in
       }
     ];
 
-    nix.settings = lib.mkIf config.catppuccin.cache.enable {
+    # Make our lives easier by centralizing the different interpretations of `enable`.
+    # That way all modules can just rely on `autoEnable` being correct, and `_enable`
+    # being a global enablement state.
+    catppuccin = {
+      # Keep `autoEnable` in sync with `enable` for releases where it didn't exist.
+      autoEnable = lib.mkIf (
+        !lib.versionAtLeast catppuccinLib.getModuleRelease "26.05"
+      ) config.catppuccin.enable;
+
+      _enable =
+        if lib.versionAtLeast catppuccinLib.getModuleRelease "26.05" then
+          config.catppuccin.enable
+        else
+          true;
+    };
+
+    nix.settings = lib.mkIf (config.catppuccin._enable && config.catppuccin.cache.enable) {
       extra-substituters = [ "https://catppuccin.cachix.org" ];
       extra-trusted-public-keys = [
         "catppuccin.cachix.org-1:noG/4HkbhJb+lUAdKrph6LaozJvAeEEZj4N732IysmU="
--- a/modules/global.nix
+++ b/modules/global.nix
@@ -1,5 +1,6 @@
 { catppuccinModules }:
 {
+  options,
   config,
   lib,
   pkgs,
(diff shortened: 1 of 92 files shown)


isabelroses:
currently unless we just break every single module. we should also check the stateVersion something roughly like `if lib.versionAtLeast catppuccinLib.getModuleRelease "26.05" then cfg.enable && config.catppuccin.enable else cfg.enable`

ThinkChaos:
Based on your other comment I thought that would be enough :/  
I'll add the `versionAtLeast` check tomorrow!

(I rebased on main)

ThinkChaos:
No worries! I think I can use an attr under `config.lib.catppuccin` to make it easier :)

isabelroses:
okay we are just kinda waiting for the sake of waiting now so i've applied some ideas i've had.

i somewhat decided that `config.catppuccin.enable` will imply `config.catppuccin.autoEnable` and this can then be disabled by users that don't like that. then we don't need `config.catppuccin._enable` and people who want auto enrolling ports disabled are happy too.

modules/global.nix
```diff
@@ -17,7 +17,17 @@ in
...
+      Catppuccin gloabl toggle.
+
+      If you intend for auto enrolling ports then you want to see
+      {option}`catppuccin.autoEnable`
+    '';
+
+    autoEnable = lib.mkEnableOption "all Catppuccin integrations by default" // {
+      default = config.catppuccin.enable;
+      defaultText = "catppuccin.enable";
+    };
```
ThinkChaos, line 46: We might as well make the default just `true` no?  
All modules are already gated by `config.catpuccin.enable`, so having a default static would be clearer IMO.

ThinkChaos (commented):
Thank you for getting back to this! :)

I see you rebased, thanks, just to be 100% sure cause I made the mistake last time, did you check if there were new modules added? Those need to be modified and don't cause conflicts so it's easy to miss!

modules/global.nix
isabelroses: i think for the sake of consistency with our other modules this is a more accurate thing to do.

but i have noticed that now break modules that do something like `catppuccin.enable = false` but have other ports enabled... i rember why we did it like this now

isabelroses:
> I see you rebased, thanks, just to be 100% sure cause I made the mistake last time, did you check if there were new modules added?

I did! thanks :D

so latest push i've had a revelation. just make the`useGlobalEnable` be `catppuccin.enable && catppuccin.autoEnable` then anything that doesn't implement that has to figure it out themselves. i've called `26.11` our release where you *must* set `catppuccin.autoEnable` and we can warn on the option not being set for now hence this logic. since if `autoEnable` is now set in a contemporary context (pre 26.11) it will still have effect.

```nix
isAutoEnrolled =
    if (lib.versionAtLeast ctp.getModuleRelease "26.11") then
      config.catppuccin.enable && config.catppuccin.autoEnable
    else if (options.catppuccin.autoEnable.highestPrio >= 1500) then
      config.catppuccin.enable && config.catppuccin.autoEnable
    else
      config.catppuccin.autoEnable;
```

isabelroses:
its a bit hot. i forgot about a logical and that could make that read a bit easier...

ThinkChaos:
> so latest push i've had a revelation. just make theuseGlobalEnable be catppuccin.enable && catppuccin.autoEnable then anything that doesn't implement that has to figure it out themselves

I had initially chosen not to do that cause it means `catppuccin.enable` has an observable effect on `catpuccin.*` options which AFAIK is not super common.  
I'm ok with it though cause my main goal here was to has a global killswitch, which is still the case.

isabelroses:
this should be good to merge now. i'd like to wait for the proper 26.05 release before i do that though

isabelroses:
had to move back to the per file `config.catppuccin.enable` since i realised my code would bypass the enable option if someone did `catppuccin.tty.enable` for example lol

ThinkChaos:
Thanks again! I switched to back mainline which is always a great feeling :)

ncfavier:
I want to enable catppuccin just for Halloy, so I did the following:

```nix
  catppuccin.enable = true;
  catppuccin.autoEnable = false;
  catppuccin.halloy = {
    enable = true;
    flavor = "mocha";
  };
```

I get this evaluation warning:

```
catppuccin/nix will soon auto enroll ports using
`catppuccin.autoEnable` and `catppuccin.enable` will act as a global
toggle to disable all ports and catppuccin modifications.

you can supress this warning by explicitly setting
`catppuccin.autoEnable` to match your current value of
`catppuccin.enable` and change `catppuccin.enable` to true
```

I don't understand the suggested action here. I don't want to set `autoEnable` to my current value of `enable` (`true`), since I don't want to automatically enable everything...

isabelroses:
@ncfavier can you run `nixos-rebuild repl` and then copy the result of `options.catppuccin.autoEnable.highestPrio` or `options.home-manager.users.valueMeta.attrs.<your username>.configuration.options.catppuccin.autoEnable.highestPrio`.

i would assume your current value might be `1500` which would explain the error

ncfavier:
`options.home-manager.users.valueMeta.attrs.n.configuration.options.catppuccin.autoEnable.highestPrio` is 
100.

isabelroses:
well thats really odd. we only check for the `autoEnable` once here:

https://github.com/catppuccin/nix/blob/8b943da8a0f8628f3446d2517ea39babcfaf27f3/modules/global.nix#L20 https://github.com/catppuccin/nix/blob/8b943da8a0f8628f3446d2517ea39babcfaf27f3/modules/global.nix#L92-L102

and you are explicitly setting the option so i cannot tell.

ncfavier:
Sorry, it was my mistake 🤦. The configuration that raised the warning indeed did not have `autoEnable` set. Thanks for your help.

PR #820, opened 2026-01-08
Respect catppuccin.fcitx5.apply

hey2022:
Simple fix to a bug introduced in #797, where the fcitx5 module no longer respects the catppuccin.fcitx5.apply setting

### DIFF
--- a/modules/home-manager/fcitx5.nix
+++ b/modules/home-manager/fcitx5.nix
@@ -41,7 +41,9 @@ in
       addons = [
         (sources.fcitx5.override { inherit (cfg) enableRounded; })
       ];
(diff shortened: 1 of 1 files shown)


isabelroses:
I think i would rather remove the option then support it again. since now you can `mkForce` it to off.

hey2022:
While you can `mkForce`, since https://github.com/nix-community/home-manager/pull/6940 any setting to fcitx5.settings will make home-manager manage `~/.config/fcitx5` preventing gui configuration through fcitx5-configtool. I feel that an option to simply disable nix catppuccin from enabling the management of the whole fcitx5 config (not just the theme now) would be convenient for people.

PR #883, opened 2026-04-05
fix(opencode): move theme to programs.opencode.tui

fredclausen:
In nix-community/home-manager@1089b2c  the home manager module was updated to support OpenCode v1.2.15+, which  requires TUI-specific settings like theme in a separate tui.json file. 

This fixes the deprecation warning about programs.opencode.settings containing TUI-specific keys.


pbek (approved):

isabelroses:
thanks!

PR #884, opened 2026-04-06
feat(home-manager): add support for hyprtoolkit

### LINKED ISSUE #779: Support Hyprtoolkit
Hyprtoolkit was just released and now allows theming:
https://wiki.hypr.land/Hypr-Ecosystem/hyprtoolkit/

**TLDR:** in `~/.config/hypr/hyprtoolkit.conf` it allows to configure:

Image

SchweGELBin:
Resolves #779

Hey, this PR adds support to hyprtoolkit as suggested in the mentioned issue.
The hyprtookit repository got added a few days ago, so I could finally add this port here.
A simple `xdg.configFile` is the best approach I could think of as a home-manager module isn't available (yet) and not really needed in this use case.

hyprtoolkit is, as its name indicates, a toolkit like gtk or qt that can be themed and preferenced by a file that's being provided by the catppuccin port.

### DIFF
--- a/modules/home-manager/hyprtoolkit.nix
+++ b/modules/home-manager/hyprtoolkit.nix
@@ -1,0 +1,21 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+
+  cfg = config.catppuccin.hyprtoolkit;
+in
+
+{
+  options.catppuccin.hyprtoolkit = catppuccinLib.mkCatppuccinOption {
+    name = "hyprtoolkit";
+    accentSupport = true;
+    useGlobalEnable = false;
+  };
+
+  config = lib.mkIf cfg.enable {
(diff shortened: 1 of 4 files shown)


modules/home-manager/hyprtoolkit.nix
```diff
@@ -0,0 +1,20 @@
...
+let
+  inherit (config.catppuccin) sources;
+
+  cfg = config.catppuccin.hyprtoolkit;
+in
+
+{
+  options.catppuccin.hyprtoolkit = catppuccinLib.mkCatppuccinOption {
+    name = "hyprtoolkit";
+    accentSupport = true;
```
isabelroses, line 13: since there is no upstream option blocking this from being enabled we should add this

```suggestion
    accentSupport = true;
    useGlobalEnable = false;
```

isabelroses (approved):

PR #885, opened 2026-04-08
Add support for Home Assistant

Aehmlo:
This PR contains a pair of commits with the goal of supporting Catppuccin theming of the Home Assistant frontend (Lovelace).

The first commit implements baseline functionality. Unfortunately, Home Assistant [doesn't provide](https://github.com/home-assistant/home-assistant.io/issues/27849#issuecomment-1678565423) any means to directly set the default frontend theme declaratively, so as a workaround, I've included an optional run-at-startup automation to set the default theme at runtime.

The second commit implements `catppuccin.home-assistant.{light,dark}Flavor` to integrate with Home Assistant's built-in light/dark theming support. This is modeled loosely on #442; the gist is that the bare `flavor` option is the default value for both dark and light flavors and each can be overridden separately. This feels likely to be forward compatible with potential future solutions for separate dark/light themes (and I'd be happy to help out with that implementation, if there's interest, separately).

I've tested this (both commits) and it works exactly as expected for me on Home Assistant 2026.4.1. I'd love to have access to it on 25.11, so I'll happily submit a backport PR if this is accepted.

### DIFF
--- a/modules/nixos/home-assistant.nix
+++ b/modules/nixos/home-assistant.nix
@@ -1,0 +1,49 @@
+{ catppuccinLib }:
+{
+  lib,
+  config,
+  ...
+}:
+let
+  inherit (config.catppuccin) sources;
+  inherit (lib) singleton toSentenceCase;
+
+  cfg = config.catppuccin.home-assistant;
+in
+{
+  options.catppuccin.home-assistant =
+    catppuccinLib.mkCatppuccinOption {
+      name = "home-assistant";
+      accentSupport = true;
+    }
+    // {
+      setDefaultAtStartup = lib.mkEnableOption "setting the default theme at Home Assistant startup" // {
+        default = true;
+        example = false;
+      };
+    };
+
+  config = lib.mkIf cfg.enable {
+    services.home-assistant.config = {
+      frontend.themes = "!include_dir_merge_named ${sources.home-assistant}";
+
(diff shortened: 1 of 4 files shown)


Aehmlo:
Added `services.home-assistant.enable = true` to NixOS tests based on previous PRs.

modules/nixos/home-assistant.nix
```diff
@@ -0,0 +1,90 @@
...
+      darkFlavor = lib.mkOption {
+        type = catppuccinLib.types.flavor;
+        default = config.catppuccin.home-assistant.flavor;
+        description = ''
+          Catppuccin flavor for Home Assistant dark mode.
+
+          Takes priority over {option}`catppuccin.home-assistant.flavor`.
+        '';
+        example = "macchiato";
+      };
```
isabelroses: i'd like to do this in one standard way at some point, https://github.com/catppuccin/nix/issues/420, so lets not include it here for now

Aehmlo:
Removed dark mode support, rebased, and squashed into a single commit. For those interested, I'll keep the dark mode patch up on [my personal fork](https://github.com/Aehmlo/catppuccin-nix/tree/home-assistant-dark) for now, until the unified approach is sorted out.

isabelroses:
thanks for the contribution!

PR #888, opened 2026-04-10
fix: replace old config scheme in opencode

OriCat101:
programs.opencode.settings contains deprecated TUI-specific keys: theme
These settings should be moved to programs.opencode.tui instead.

OpenCode v1.2.15+ requires TUI settings in a separate tui.json file. See: https://opencode.ai/docs/config#tui


OriCat101:
ups, should have checked open PR's #883

PR #891, opened 2026-04-14
fix(neovim): make config type explicit

musjj:
The default type of config language now depends on `home.stateVersion`:

```nix
programs.neovim.plugins.*.type
    Language used in config. Configurations are aggregated per-language.

    Type: one of “lua”, “viml”, “teal”, “fennel” or string

    Default:

        if lib.versionAtLeast config.home.stateVersion "26.05" then "lua" else "viml"
```

So if you run catppuccin in a `26.05` environment, the current config will break:

```
E5112: Lua chunk: /home/$USER/.config/nvim/init.lua:6: '=' expected near '<'
```

This PR changes it so that the config is explicitly typed as lua instead.

### DIFF
--- a/modules/home-manager/neovim.nix
+++ b/modules/home-manager/neovim.nix
@@ -23,16 +23,15 @@ in
       plugins = [
         {
           plugin = config.catppuccin.sources.nvim;
+          type = "lua";
           config = ''
-            lua << EOF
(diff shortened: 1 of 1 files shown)


isabelroses:
thanks!

PR #902, opened 2026-04-28
ci: checkout with default token

backwardspy:
don't think this token is actually needed, unless i've missed something. none of the other workflows use one. default is `github.token` which should be enough unless we need to clone private submodules etc.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -16,8 +16,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


sgoudham:
The default token won't be able to bypass the ruleset restrictions on the main branch when it tries to push back, so we should kill the format workflow or be okay with it failing once a year when I inevitably forget to renew my token.

sgoudham:
Oh wait, it's not pushing back? Lmfao

sgoudham (approved):

PR #910, opened 2026-04-30
fix(home-manager/wezterm): add settings & handle programs.wezterm.settings

### LINKED ISSUE #909: Wezterm theme does not apply without extraConfig
Likely manifesting from https://github.com/nix-community/home-manager/commit/92d382b9821aad4036b9b7e4270a959f24909a67, there is no longer a strict need to use `programs.wezterm.extraConfig`. The current module  assumes `config` will be returned at the end of `extraConfig`.

Example config:

```nix
programs.wezterm = {
  enable = true;

  settings = {
    hide_tab_bar_if_only_one_tab = true;
    window_padding = {
      left = 30;
      right = 30;
      top = 30;
      bottom = 30;
    };
  };
};

catppuccin.wezterm = {
  enable = true;
  apply = true;
};
```
This will not work, without adding
```nix
extraConfig = ''
    return config
'';
```

isabelroses:
here i've copied the settings concept that we have in catppuccin.nvim. and i've essentially soft deprecated `apply` for people who use `programs.wezterm.settings` and i think this is a much better use. maybe we should also warn here too.

closes #909 

@kitgxrl whats your thoughts? :D

### DIFF
--- a/modules/home-manager/wezterm.nix
+++ b/modules/home-manager/wezterm.nix
@@ -3,6 +3,10 @@
 let
   inherit (config.catppuccin) sources;
   cfg = config.catppuccin.wezterm;
+
+  defaultConfig = {
+    inherit (cfg) flavor accent;
+  };
 in
 {
   options.catppuccin.wezterm =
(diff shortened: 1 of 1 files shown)


kitgxrl:
Works great for me! Appreciate the fast fix

PR #917, opened 2026-05-05
feat(home-manager/kvantum): use qt.kvantum options

name-tar-xz:
Uses the kvantum module present in home-manager instead of overwriting config files.
It currently needs `kvantum.enable` (upstream PR to fix that https://github.com/nix-community/home-manager/pull/9269) 
and it adds every theme's files into `.config/Kvantum`

Also fixed a typo.

### DIFF
--- a/modules/home-manager/kvantum.nix
+++ b/modules/home-manager/kvantum.nix
@@ -19,21 +19,12 @@ in
       accentSupport = true;
     }
     // {
-      apply = lib.mkOption {
-        type = lib.types.bool;
-        default = true;
-        description = ''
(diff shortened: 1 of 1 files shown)


isabelroses:
feel free to mark as ready for review when the upstream pr is merged

musjj:
This doesn't really require the upstream PR no? The module already exists, the upstream PR just changes the default for `enable`.

isabelroses:
we would want to move `config.qt.enable` to `config.kvantum.enable` however so i don't quite agree about it not needing upstream

PR #928, opened 2026-05-13
fix(plymouth): fix path in existing themes files

mjm:
this fix should allow the plymouth theme to work again, which it doesn't currently.

the plymouth package is not performing a build. instead it is just copying the existing built themes from the source repo. however, the code to fix the ImageDir paths in the themes was changed to operate on the source template instead of the built themes like it used to. this change reverts that.

perhaps a better change would be to actually perform the build here in this package, but that is a more involved change due to the images needing to be converted to pngs. for now, this is the simple fix to get the theme working again.


isabelroses (approved):
thanks

PR #934, opened 2026-05-18
fix(home-manager/kvantum): use nix option instead of writing file

PerchunPak:
For some reason, this fixes `~/.config/Kvantum/kvantum.kvconfig` for me


isabelroses:
duplicate of https://github.com/catppuccin/nix/pull/917. its basically just not merged yet since this is effectively a breaking change and were prepping for the next release. without the changes to the enable option in https://github.com/nix-community/home-manager/pull/9269

PR #936, opened 2026-05-18
fix(vscode): correct vendor hash

isabelroses:


isabelroses:
ahh i hate this fetcherrrrr

PR #939, opened 2026-05-23
feat(home-manager): add support for obsidian

nickthegroot:
image


modules/tests/home.nix
```diff
@@ -80,6 +80,7 @@ in
     neovim.enable = true;
     newsboat.enable = true;
     obs-studio.enable = isLinux;
+    obsidian.enable = true;
```
nickthegroot, line 83: ❓ Obsidian is an unfree package. Would y'all prefer disabling the test, or allowing building of unfree packages?

PR #941, opened 2026-05-23
feat(home-manager): add support for obsidian

nickthegroot:
Fixed version of #939

Note that `obsidian` is listed as an `unfree` package, thus requiring the changes to `dev/flake.nix`. There are other options that I'd be happy to switch to if desired:

1. Use the unfree package filter (`nixpkgs.config.allowUnfreePredicate`)
2. Disable testing for `obsidian`

### DIFF
--- a/modules/home-manager/all-modules.nix
+++ b/modules/home-manager/all-modules.nix
@@ -47,6 +47,7 @@
   ./newsboat.nix
   ./nushell.nix
   ./obs.nix
+  ./obsidian.nix
   ./opencode.nix
   ./polybar.nix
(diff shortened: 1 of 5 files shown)


isabelroses (commented):
i think its alright. i just don't think we should allow unfree packages in ci.

nickthegroot:
> i think its alright. i just don't think we should allow unfree packages in ci.

Fair enough! I reverted those changes, and added a note to the disabled test. Let me know if you have any other changes.

isabelroses (approved):

PR #948, opened 2026-05-26
feat(home-manager/gemini-cli): remove IFD

isabelroses:
https://geminicli.com/docs/cli/themes/#loading-themes-from-a-file

### DIFF
--- a/modules/home-manager/gemini-cli.nix
+++ b/modules/home-manager/gemini-cli.nix
@@ -5,7 +5,6 @@ let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.gemini-cli;
-  theme = lib.importJSON "${sources.gemini-cli}/catppuccin-${cfg.flavor}.json";
 in
 
 {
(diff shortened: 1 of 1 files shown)


SymphonySimper:
This broke Gemini CLI's theming. This was already brought up in the [previous PR](https://github.com/catppuccin/nix/pull/827#discussion_r2698917436). So please consider reverting the change.

Note from official docs for reference ([source](https://geminicli.com/docs/cli/themes/#loading-themes-from-a-file:~:text=For%20your%20safety,from%20untrusted%20sources.)):

> Caution
> For your safety, Gemini CLI will only load theme files that are located within your home directory. If you attempt to load a theme from outside your home directory, a warning will be displayed and the theme will not be loaded. This is to prevent loading potentially malicious theme files from untrusted sources.

SymphonySimper:
The support can also be removed entirely, as it's going to be discontinued in favor of the Antigravity CLI by June 18 anyway.

Source: https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/

isabelroses:
> Caution
For your safety, Gemini CLI will only load theme files that are located within your home directory. If you attempt to load a theme from outside your home directory, a warning will be displayed and the theme will not be loaded. This is to prevent loading potentially malicious theme files from untrusted sources.

we could just use `xdg.configFile` then no biggie

isabelroses:
unlessss they don't use xdg because ai slop right `home.file` it is

PR #961, opened 2026-05-30
fix(home-manager/gemini-cli): use home.file

isabelroses:
see https://github.com/catppuccin/nix/pull/948#issuecomment-4582755309


SymphonySimper:
Same thing, it doesn't even allow symlinks that are outside of home. I pretty much tried everything when I implemented the theming with IFD.

image
image

isabelroses:
amazing

SymphonySimper:
> amazing

Yep -_-

PR #975, opened 2026-06-06
feat!(home-manager/gemini-cli): remove

### LINKED ISSUE #974: Replace programs.gemini-cli.settings with programs.antigravity-cli.settings
Now `home-manager switch` outputs:
```
warning: The option `programs.gemini-cli.settings' defined in `/nix/store/6ar2vjfzlhxwqqzhd8sjxkjp9dyh9s78-source/modules/home-manager/gemini-cli.nix' has been renamed to `programs.antigravity-cli.settings'
```

isabelroses:
closes https://github.com/catppuccin/nix/issues/974 

@SymphonySimper as far as i could tell from the [docs](https://antigravity.google/docs/cli-reference) the only option is `colorScheme` which doesn't support custom keys options.

### DIFF
--- a/dev/flake.lock
+++ b/dev/flake.lock
@@ -39,11 +39,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1780593650,
-        "narHash": "sha256-CHo7k65YTL3HY+WQVedDTupji+LMgNlKCdrtRHZFAK4=",
+        "lastModified": 1780679734,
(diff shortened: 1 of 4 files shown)


SymphonySimper:
Yep, seems like they completely removed the custom theme option. I tested it with `colorScheme` set to `terminal` and it works without any issues.

PR #976, opened 2026-06-06
feat(home-manager/fzf): use fzfrc file

isabelroses:
depends on https://github.com/catppuccin/fzf/pull/30. removes IFD

### DIFF
--- a/modules/home-manager/fzf.nix
+++ b/modules/home-manager/fzf.nix
@@ -1,42 +1,35 @@
 { catppuccinLib }:
-{ config, lib, ... }:
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
 
 let
   inherit (config.catppuccin) sources;
 
   cfg = config.catppuccin.fzf;
-  palette = (lib.importJSON "${sources.palette}/palette.json").${cfg.flavor}.colors;
-
-  # Manually populate with colors from catppuccin/fzf
-  # The ordering is meant to match the order of catppuccin/fzf to make comparison easier
-  colors = lib.attrsets.mapAttrs (_: color: palette.${color}.hex) {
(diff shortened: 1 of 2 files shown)


budimanjojo:
Thank you for the effort of removing IFD for FZF!

Looking at the commit, my assumption is that this will require the users to manually set `programs.fzf.colors`?
If so, I hope you can consider manually setting the colors here instead. I don't think the upstream will change the color palette anyway (palette.json file didn't have any changes at all since it was first created other than bumping version and adding color format support).

budimanjojo:
I have another suggestion, what about a GitHub Actions workflow that do a regular pulling of upstream palette.json file into this repo and convert it into a `.nix` file so we can consume it locally here?

isabelroses:
> my assumption is that this will require the users to manually set programs.fzf.colors

no. it functions the exact same as before. a users will only need to set these if they intend to override the colors.

> I have another suggestion, what about a GitHub Actions workflow that do a regular pulling of upstream palette.json file into this repo and convert it into a .nix file so we can consume it locally here?

this exists in another pr https://github.com/catppuccin/nix/pull/738.

sollniss:
FYI, it says "a accent" in the description of the removed option.

talal:
@isabelroses can you please backport this to 26.05 so that we don't have to wait until November?

This change relies on fzf v0.47.0 which was released in 2024 so this shouldn't break anybody's setup 🤞

isabelroses:
@talal i was going to backport but i've just realised this PR removes accent support which i would consider a breaking change. so i don't think i can backport this until we readd accent support

talal:
@isabelroses I see, no worries. Thanks for the quick response :)

PR #996, opened 2026-07-03
feat(home-manager): add support for clipse

saadndm:
Adds support for https://github.com/catppuccin/clipse

### DIFF
--- a/modules/home-manager/clipse.nix
+++ b/modules/home-manager/clipse.nix
@@ -1,0 +1,21 @@
+{ catppuccinLib }:
+{ config, lib, ... }:
+
+let
+  inherit (config.catppuccin) sources;
+  cfg = config.catppuccin.clipse;
+  enable = config.catppuccin.enable && cfg.enable && config.services.clipse.enable;
+in
+
+{
+  options.catppuccin.clipse = catppuccinLib.mkCatppuccinOption {
+    name = "clipse";
+    accentSupport = true;
+  };
+
+  config = lib.mkIf enable {
+    services.clipse = {
+      theme = sources.clipse + "/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.json";
+    };
+  };
+}
--- a/modules/home-manager/all-modules.nix
+++ b/modules/home-manager/all-modules.nix
@@ -10,6 +10,7 @@
   ./btop.nix
   ./cava.nix
   ./chrome.nix
+  ./clipse.nix
   ./cursors.nix
   ./delta.nix
   ./dunst.nix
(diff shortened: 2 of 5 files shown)


isabelroses:
this pr is going to need some changes now that https://github.com/nix-community/home-manager/pull/9591 was merged

modules/home-manager/clipse.nix
```diff
@@ -0,0 +1,21 @@
...
+
+{
+  options.catppuccin.clipse = catppuccinLib.mkCatppuccinOption {
+    name = "clipse";
+    accentSupport = true;
+  };
+
+  config = lib.mkIf enable {
+    services.clipse = {
+      theme = lib.importJSON "${sources.clipse}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.json";
```
isabelroses:
```suggestion
      theme = "${sources.clipse}/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.json";
```

saadndm: Thanks for improving the HM module, but theme is now typed as:
```nix
lib.types.either jsonFormat.type lib.types.path
```
but Catppuccin needs to pass a store path from a derivation, and I don't know of a way to clean way to convert a store path string to a path type? (see NixOS/nix#12407)

Should we continue using `lib.importJSON`, or could `theme` also accept `lib.types.str` in the HM module?

isabelroses: you could adapt it to `sources.clipse + "/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.json"` this would mean its a path

edit: ah you found that

saadndm:
@isabelroses But when testing locally `sources.clipse + "/${cfg.flavor}/catppuccin-${cfg.flavor}-${cfg.accent}.json"` didn't work, because `sources.clipse` is a derivation which gets converted to a string, so my `~/.config/clipse/custom_theme.json` becomes a store path string:
```json
"/nix/store/0klql8mi22n5hm6r0h26mavnwmdql2xf-catppuccin-mocha-blue.json"
```
instead of the actual theme contents.

I should have made it clear in this PR that my changes didn't work, sorry for the confusion.

It's currently broken, I think the best fix is for the clipse home-manager module to check `lib.isStorePath` (https://nixos.org/manual/nixpkgs/stable/#function-library-lib.strings.isStorePath) like this:
```nix
        if lib.isPath cfg.theme || lib.isStorePath cfg.theme then
          cfg.theme
        else
          jsonFormat.generate "theme" cfg.theme;
```

From what I can see, many other packages like btop, helix, gitui etc. also do this.

isabelroses:
seems good. would you be willing to upstream that?

saadndm:
Sure, I just did in https://github.com/nix-community/home-manager/pull/9612

PR #1000, opened 2026-07-08
feat(home-manager/cursors): support global enable

isabelroses:

PR #1003, opened 2026-07-08
feat(home-manager/cursors): support global enable v2

fredclausen:
Latest home manager commit [1aac8895](https://github.com/nix-community/home-manager/commit/1aac8895fea6fcabc6a0184c63a80fb2f99fd208) changed the default behavior of `home.pointerCursor.enable`, and now requires `home.pointerCursor.enable = true` to be explicitly set.

Below is the error message generated

```shell
evaluation warning: fred profile: Relying on `home.pointerCursor` to enable cursor config generation is deprecated.
Please update your configuration to explicitly set:

home.pointerCursor.enable = true;
```

Additionally, the test module required a small change to assert the cursor only on linux platforms, because upstream [home-cursor.nix](https://github.com/nix-community/home-manager/blob/1aac8895fea6fcabc6a0184c63a80fb2f99fd208/modules/config/home-cursor.nix#L177) asserts this is only a valid option on linux.

### DIFF
--- a/modules/home-manager/cursors.nix
+++ b/modules/home-manager/cursors.nix
@@ -2,6 +2,7 @@
 {
   options,
   config,
+  pkgs,
   lib,
   ...
 }:
@@ -35,9 +36,7 @@ in
   options.catppuccin.cursors =
     catppuccinLib.mkCatppuccinOption {
       name = "pointer cursors";
-      # NOTE: We exclude this as there is no `enable` option in the upstream
-      # module to guard it
-      useGlobalEnable = false;
+      useGlobalEnable = pkgs.stdenv.hostPlatform.isLinux;
     }
     // {
       accent = lib.mkOption {
--- a/modules/tests/home.nix
+++ b/modules/tests/home.nix
@@ -20,6 +20,9 @@ in
     hyprtoolkit.enable = true;
     xfce4-terminal.enable = isLinux;
 
+    # `home.pointerCursor` asserts the Linux platform, so only enable there
+    cursors.enable = isLinux;
+
     # we install the themes but don't apply them so we can test kvantum, qt5ct
     # and qt6ct
     kvantum.assertStyle = false;


modules/home-manager/cursors.nix
```diff
@@ -49,6 +49,7 @@ in
 
   config = lib.mkIf enable {
     home.pointerCursor = {
+      enable = true;
```
isabelroses: we shouldn't adjust users configs like this. you should have to set this yourself.
```suggestion
```

modules/home-manager/cursors.nix
isabelroses:
```suggestion
      useGlobalEnable = pkgs.stdenv.hostPlatform.isLinux;
```

isabelroses (commented):
see also my pior work in https://github.com/catppuccin/nix/pull/1000

modules/home-manager/cursors.nix
fredclausen: Thank you for taking the time to review. Before I update the PR, I want to resolve one tension — my original enable = true was the wrong way to address it, but the underlying issue is real.

Without an explicit enable key in home.pointerCursor, home-manager emits a deprecation warning. The check in home-cursor.nix is:

```nix
warnings = optional (pointerCursorDefined && !enableDefined) "...";
```

Setting name/package makes pointerCursorDefined = true, and omitting enable makes enableDefined = false — so the warning fires. The only way to silence it is to include an enable key.

Would this be the right fix, following the "theme but don't force-enable" pattern?

```nix
enable = lib.mkDefault false;
```

That gives an explicit enable key (silences the warning), keeps cursor generation off until the user enables home.pointerCursor themselves — matching how the other modules gate on the underlying programs.<foo>.enable — and because it's mkDefault, a user's own home.pointerCursor.enable = true overrides it cleanly rather than conflicting.

PR #1005, opened 2026-07-08
fix(home-manager/cursors): v3

isabelroses:


isabelroses:
yeahhhh right lol

isabelroses:
sigh... ill figure this out later

fredclausen:
I did a quick test of this on my local machine, and I think the fix may be

```nix
home.pointerCursor.enable = lib.mkDefault false;
```

And revert to

```nix
useGlobalEnable = pkgs.stdenv.hostPlatform.isLinux;
```

This silences any deprecation warnings by setting the enable key with a sane value that doesn't turn on cursors unless the user specifically opts in (sorry my previous PR got that wrong), and doesn't get in a weird infinite recursion trap.

If I'm off base, I apologize and I'll leave you be.

isabelroses:
you see we cannot set `home.pointerCursor.enable = lib.mkDefault false;` since it would not work / break if someone has set `home.pointerCursor = null` which is still a valid configuration for the time being.

PR #1018, opened 2026-07-18
fix(foot): use the correct config path

nonsymplectic:
Repo structure in catpppuccin/foot has changed:
themes/catppuccin-*.ini: dark themes only, with support for light/dark switching
themes/static/catppuccin-*.ini: all themes, in legacy format without light/dark switching

Current behavior breaks the latte flavor for foot.


isabelroses:
Thanks

PR #1023, opened 2026-07-26
home-manager/cursors: explicitly set home.pointerCursor.enable

DerGrumpf:
Fixes the following home-manager evaluation warning:

```
evaluation warning: <user> profile: Relying on `home.pointerCursor` to enable cursor config generation is deprecated.
                    Please update your configuration to explicitly set:
                      home.pointerCursor.enable = true;
```

This sets \`home.pointerCursor.enable = true\` explicitly wherever the module previously relied on implicit enabling.

### DIFF
--- a/modules/home-manager/cursors.nix
+++ b/modules/home-manager/cursors.nix
@@ -48,6 +48,7 @@ in
 
   config = lib.mkIf enable {
     home.pointerCursor = {
+      enable = true;
       name = "catppuccin-${cfg.flavor}-${cfg.accent}-cursors";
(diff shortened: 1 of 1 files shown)


Daholli:
Thanks for the PR, I was just about to open the same one :D

DerGrumpf:
No Problem just merge it. Its just one line. xD

isabelroses:
theres been 3 or so prs like this! we cannot do this. set this option yourself. we will move cursors to the global enable using the value of `home.pointerCursor.enable`, which will make it so that if we set `home.pointerCursor.enable` in catppuccin/nix it will have no effect.

